PR #43, opened 2025-10-29
Remove cluster from context based on current route

torchiaf:
- When navigating from cluster explorer to home, the `rootGetters['currentCluster']` in the store still keeps the old cluster name, so it remains in the Chat context. Comparing it with the current path is the easiest way to remove the cluster from the context store ATM.


rak-phillip (approved):
LGTM

PR #47, opened 2025-10-29
Update styling of chat area

rak-phillip:
This updates the styling of the chat area.

## Screenshots

image

### DIFF
--- a/pkg/rancher-ai-ui/components/panels/Console.vue
+++ b/pkg/rancher-ai-ui/components/panels/Console.vue
@@ -157,6 +157,7 @@ watch(() => text.value, () => {
   overflow: hidden;
   text-overflow: ellipsis;
   min-width: 0;
+  font-size: 12px;
 }
 
 .chat-console-row {


pkg/rancher-ai-ui/components/message/Source.vue
```diff
@@ -27,8 +28,20 @@ const isCollapsed = ref(true);
...
       class="chat-msg-source-tags"
     >
-      <span class="chat-msg-source-tag">Cluster Management Guide</span>
-      <span class="chat-msg-source-tag">Best Practices</span>
+      <context-tag
+        :remove-enabled="false"
+        type="user"
+        class="chat-msg-user-context-tag"
+      >
+        Cluster Management Guide
```
torchiaf: These labels are not visible ATM, I think it's just because is a WIP . Other changes LG

Image

rak-phillip:
@oboc-sts thanks for the review:

> In the "context" area (right above the textarea), the distance between the "tags" and the top and bottom borders is still not 16px

It looks like the small button for the context trigger had more height than the tags. I've made an update so that the button and tags are the same height. This should be fixed now.
 
> The left-padding used on the small chips for the "source" and "context" under the chat bubbles is not the right one (pun intended)

This looks to be an empty icon tag from the ContextTag component. I've removed the option for rendering icons altogether - it looks like we don't want any icons in tags.

> The icons on the context chips / tags are still there - it might be that their removal is not part of the current PR though

This has been removed from ContextTag.

PR #49, opened 2025-10-30
Update styles for code blocks in chat messages

rak-phillip:
This updates the styles for code blocks in chat messages. The line height of text in chat messages is also updated.

## Details

This addresses the following comments:

- For inline `<code>` blocks (and only for the inline `<code>` blocks), we need to get rid of the paddings, backgrounds, borders, etc.
- Keep the monospaced font and change the text colour to another colour (green?)
- The line-height used in bubbles  (both user and assistant bubbles) however is something we really must fix - increase it to 21px please

We currently don't implement the following: 

- the user can interact with (e.g. on hover I get the copy-to-clipboard icon that lets me copy the code)

I think that we will require a larger change to get this. We have this feature elsewhere, but it's implemented specifically for the `KeyValueRow` component. 

## Screenshots

image

image

### DIFF
--- a/pkg/rancher-ai-ui/components/message/index.vue
+++ b/pkg/rancher-ai-ui/components/message/index.vue
@@ -291,6 +291,7 @@ onBeforeUnmount(() => {
   box-shadow: 0 2px 8px 0 rgba(0,0,0,0.04);
   padding: 12px;
   font-size: 0.95rem;
+  line-height: 21px;
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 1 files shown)


oboc-sts:
The update on the inline code blocks looks delicious @rak-phillip  💪 

The IxD (hover to copy the code to clipboard) is NOT intended for these inline blocks, only for the standalone block blocks - so indeed, it's out of scope for this ticket.

PR #51, opened 2025-10-31
Implement Chat processing label

torchiaf:
[Screencast from 2025-10-31 13-59-08.webm](https://github.com/user-attachments/assets/f3652362-2dda-46bc-b0ff-ad9412ae8a41)


rak-phillip (approved):
LGTM - we just need to rebase first. I was considering if the message should be floating to prevent shifting elements. I don't think this is a concern, given the nature of the chat.

PR #55, opened 2025-10-31
Update the warning for third-party providers

rak-phillip:
This updates the warning for third-party providers on the settings page so that it is more explicit and clear.


torchiaf (approved):
LGTM

PR #56, opened 2025-10-31
Render welcome message immediately

rak-phillip:
This immediately renders the welcome message, independently of other messages.

## Video

https://github.com/user-attachments/assets/6ae4ea05-0bd9-46de-9c1a-fae75734d498


rak-phillip:
@oboc-sts This changes the order of the welcome messages. This differs from the original design, but it is done to keep the flow of the chat messages looking natural.

torchiaf (changes requested):
We worked on the same issue without knowing it :satisfied: , sorry, yours is older.
However, I think this one has less impact, because it only modifies a few input props and keeps the possibilities to have different message templates: https://github.com/rancher-sandbox/rancher-ai-ui/pull/57/commits/645760b62ae3808635b1864efbfcbe94287b5420
so I would vote to merge it.

I changed the order of the messages too, to display the suggestions at the bottom.

Let me know, I'm open to revert my commit.

PR #75, opened 2025-11-05
Destroy overlays when clicking open button

torchiaf:


rak-phillip (approved):
LGTM

torchiaf:
@rak-phillip thanks for the review, I pushed a new commit.
Anyway, we'll have to do more tests around it, so let's wait before merging.

PR #77, opened 2025-11-05
Update condition for rendering privacy notice banner

rak-phillip:
The privacy notice banner is using an outdated value to conditionally render. This updates it so that it uses the enum instead of the string "Local".


torchiaf (approved):
LGTM

PR #88, opened 2025-11-19
Fix Chat.isOpen

### LINKED ISSUE #86: Rancher AI chat window cannot open when another window occupies right position
Related issue: https://github.com/rancher/dashboard/issues/15941

torchiaf:
Fixes #86 

- wm/isOpen is not working as expected. We need to fix this logic in the extension and restore wm.isOpen when fixed in the shell.


rak-phillip (approved):
LGTM

PR #96, opened 2025-12-01
Bedrock integration

jp-gouin:
Blocked by https://github.com/rancher-sandbox/rancher-ai-agent/pull/45

Add the Bedrock tile to work with Bedrock integration 
https://github.com/rancher-sandbox/rancher-ai-ui/issues/95

This PR add the support of the Bedrock and the configuration of 
`AWS_SECRET_ACCESS_KEY` : The AWS secret key
`AWS_ACCESS_KEY_ID ` : The AWS access key
`AWS_BEARER_TOKEN_BEDROCK ` : Optional the Bedrock token for short or long term personal token support
`AWS_REGION_NAME ` : The AWS region

The PR also include several changes
- Change the default local model to the recommended one (`gpt-oss:20b`)
- Fix a bug with the default model when switching providers
   The form always defaulted to the `models` when switching provider ignoring the one in the formData if set with the agent helm deployment
- Allow users to type their own local/bedrock model instead of needing to update the helm chart
- Added new bedrock icon

### DIFF
--- a/pkg/rancher-ai-ui/components/panels/Console.vue
+++ b/pkg/rancher-ai-ui/components/panels/Console.vue
@@ -133,7 +133,10 @@ watch(() => text.value, () => {
       </div>
     </div>
     <div class="chat-console-row chat-console-chat-text-info">
-      <span class="chat-model label text-deemphasized">
+      <span
+        v-clean-tooltip="!!props.agent && props.agent.model?.length > 20 ? props.agent.model : ''"
+        class="chat-model label text-deemphasized"
+      >
         {{ !!props.agent ? t('ai.agent.label', { name: props.agent.name, model: props.agent.model }, true) : t('ai.agent.unknown') }}
       </span>
       <TextLabelPopover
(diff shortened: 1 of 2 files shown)


rak-phillip:
> TODO: Add the icon, I tried to use icon-aws but it does not exist in the icon library and I don't know how to add it :)

@kwwii @oboc-sts I created a new issue in the Dashboard repo to add a new icon rancher/dashboard#16089

torchiaf (changes requested):
I left some comments, I think we should:

- eventually remove this code:
```
  // Initialize the models based on the formData values
  if (formData.value[Settings.ACTIVE_CHATBOT]) {
    models[formData.value[Settings.ACTIVE_CHATBOT] as ChatBotEnum] = formData.value[Settings.MODEL]
      ? [formData.value[Settings.MODEL]]
      : models[formData.value[Settings.ACTIVE_CHATBOT] as ChatBotEnum];
  }
```

- add an icon for Bedrock, this can be done bumping @rancher/icon when the new version is available

- lint the code: `yarn lint --fix`

PR #103, opened 2025-12-15
Wire @rancher/cypress

torchiaf:
Contributes to https://github.com/rancher-sandbox/rancher-ai-ui/issues/109


pkg/rancher-ai-ui/pages/settings/Settings.vue
```diff
@@ -144,7 +144,12 @@ const resource = useFetch(async() => {
         }
       });
 
-      data.save();
+      try {
```
torchiaf: We are adding the try-catch fir safeness.
The Settings page shouldn't be visible if the Agent is not installed - this will be addressed in the future

rak-phillip (approved):
LGTM - This looks much improved with the inclusion of the new `@rancher/cypress` package.

Issue #117, opened 2026-01-12
manual agent selection


samjustus:
User should be able to select the agent manually.
When switching agent, the conversation should be kept.

backend ticket https://github.com/rancher-sandbox/rancher-ai-agent/issues/62

PR #124, opened 2026-01-29
Multi agent support

torchiaf:
Fixes https://github.com/rancher/rancher-ai-ui/issues/117

This will add the possibility to select specific agents or the adaptive selection in the console.

- It's possible to use the adaptive mode - the agent is automatically selected by the backend
- It's possible to manually select an agent
- It's possible to manually select the agent as suggested by the system

## TODO

- [x] Code rebase
- [x] Rename the agent's name on each message so that it will be meaningful for the users: "Fleet" -> "Fleet Agent" 
- [x] Fix agent info label - it should be aligned to the left

[Screencast from 2026-02-09 12-13-49.webm](https://github.com/user-attachments/assets/71abfc12-68e1-4f16-ab0b-c8ef8bc735d9)

### DIFF
--- a/pkg/rancher-ai-ui/composables/useAIServiceComposable.ts
+++ b/pkg/rancher-ai-ui/composables/useAIServiceComposable.ts
@@ -94,6 +94,8 @@ export function useAIServiceComposable() {
   async function checkAgentDeploymentAvailability() {
     if (!store.getters['management/canList'](WORKLOAD_TYPES.DEPLOYMENT)) {
       error.value = { key: 'ai.error.services.deployment.noPermission' };
+
+      return false;
     } else {
       try {
         const agent = await store.dispatch('management/find', {


pkg/rancher-ai-ui/composables/useAIServiceComposable.ts
```diff
@@ -0,0 +1,179 @@
...
+      };
+    }
+
+    return null;
+  }
+
+  async function checkAgentDeploymentAvailability() {
+    if (!store.getters['management/canList'](WORKLOAD_TYPES.DEPLOYMENT)) {
+      error.value = { key: 'ai.error.services.deployment.noPermission' };
+    } else {
```
rak-phillip: nit: Similar to `getLLMConfigs()` below, this condition can act like a guard and return early.

rak-phillip (approved):
LGTM

PR #135, opened 2026-02-12
Look & feel improvements

torchiaf:
We are fixing some mis-alignments between the UI and the original desing

In the order of commits:

- Fix the disabled color in the settings - this was caused by a global css class in the chat panel

  Before
  
  image
  
  After
  
  image

- Fix suggestions

  Before
  
  image
  
  
  After
  
  image

- Fix context's tags and show the tags and console in a single panel

  Before
  
  image
  
  After
  
  image

- Fix Result's disclaimer - some minor fixes + opacity 95%

- Fix the bubble-buttons - we are increasing the top padding to show the new buttons

  Before
  
  
  image

  
  After
  
  image
  
- Add System avatar to Agent's system suggestion message

  Before
   
  image

  After

  image

- Fix autoscroll -> when loading an old chat, we need to scroll down the list of messages and hide the ScrollButton

### DIFF
--- a/pkg/rancher-ai-ui/components/panels/Messages.vue
+++ b/pkg/rancher-ai-ui/components/panels/Messages.vue
@@ -24,6 +24,10 @@ const store = useStore();
 const { t } = useI18n(store);
 
 const props = defineProps({
+  activeChatId: {
+    type:    String,
+    default: '',
+  },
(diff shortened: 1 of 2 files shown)


rak-phillip (approved):
LGTM

PR #137, opened 2026-02-16
Multi-agent validations

torchiaf:
Contributes to https://github.com/rancher/rancher-ai-agent/issues/57

We implemented the UI logic for display information and managing the AI Agents status.

- Settings page:

  - The Agent's status is now visible in the Tab's label and in the error banner on top of each Agent's form. The tooltip shows validation errors if any, otherwise it shows the agent's status.
  
    image

  - it display an error message if there are no active agents (enabled + active status)

    image

- Chat

  - When an Agent is not available, it shows up as disabled
  
      image

  - When only 1 Agent is available, we don't show the Adaptive selection
  
    image

  - When all agents in the system are not available we show an error and the Chat is disabled

    image

### DIFF
--- a/cypress/e2e/blueprints/aiAgentConfigs.ts
+++ b/cypress/e2e/blueprints/aiAgentConfigs.ts
@@ -16,3 +16,22 @@ export const harvesterAgentConfig = {
     toolSet:            'harvester'
   }
 };
+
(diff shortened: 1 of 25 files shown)


rak-phillip (approved):
LGTM

Issue #98, opened 2025-12-10
Integrate provider settings validation with error handling


jp-gouin:
## Feature

Update the UI to validate provider settings before applying them:

- Call the agent's new provider settings validation endpoint before finalizing any provider settings (Local, Gemini, OpenAI, Bedrock).
- If validation fails, display a comprehensive error message to the user explaining the cause and which provider or setting is invalid.
- Block changes until all issues are resolved.

### Acceptance Criteria
- UI invokes the validation endpoint before saving settings.
- On validation failure, show a clear, actionable error in the UI (specifying provider and failed setting).
- Ensure error state and user feedback are thoroughly tested.

Issue #140, opened 2026-02-23
RBAC - Split llm-config


jp-gouin:
end 2 end issue for improving RBAC management on Liz.

We should split the llm-secret into two secrets. 
One containing only the LLM provider API Key that should be be accessible by regular Liz user.
One containing LLM information (model, url, provider) that can be accessible by regular Liz user to improve the UX.

PR #144, opened 2026-02-25
Settings validation

torchiaf:
Fixes https://github.com/rancher/rancher-ai-ui/issues/98
Contributes to https://github.com/rancher/rancher-ai-ui/issues/140
UX issue: https://github.com/rancher/ux/issues/391

  - We implemented the settings validation in the AI settings page.
  - The UI now support the settings source split into ConfigMap/Secret to be able to handle permissions granularity
  - The UI is also able to fetch the list of available models based on the selected llm provider:
    - Ollama: a valid Ollama URL is required to fetch the list of model from the Ollama server
    - OpenAi, Gemini: the list of models is based on what supported by the backend (hardcoded list)
    - Bedrock: it requires a valid {region, beareToken} to be fetched
  - The Apply button is now disable when we have UI or API validation errors.
  - We also removed some unused settings keys - (i.e. Bedrock legacy token and ApiKey).

[Screencast from 2026-03-02 18-59-21.webm](https://github.com/user-attachments/assets/d5e8dee1-3518-4be1-9f06-8030bd60dc81)

### DIFF
--- a/pkg/rancher-ai-ui/l10n/en-us.yaml
+++ b/pkg/rancher-ai-ui/l10n/en-us.yaml
@@ -330,10 +330,10 @@ aiConfig:
       description: The AI model to use for generating responses
     OLLAMA_MODEL:
       label: Model
-      description: 'The AI model list fetched from your Ollama instance. Make sure to add a <b>valid Ollama URL</b> to fetch the models or try to <refreshModels>refresh</refreshModels>.'
+      description: 'Make sure you provide a <b>valid Ollama URL</b> to fetch the available models. <refreshModels>Refresh now</refreshModels>'
     BEDROCK_MODEL:
       label: Model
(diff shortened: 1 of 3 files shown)


cypress/e2e/tests/features/settings.spec.ts
```diff
@@ -91,20 +91,16 @@ describe('AI Assistant Configuration', () => {
 
       aiAgentConfigs.checkExists();
 
-      // Check that both Rancher Agent and Fleet Agent tabs are present
```
torchiaf: The Agent is now installed without Fleet built-in agent.

oboc-sts (commented):
Everything looks fine and according to design with a few exceptions.

During the review, we've updated:
- error messages
- models description

And we've noticed a **small bug** (screenshot below) - the error displayed there is visible before the user touched the required field for bedrock

Image

Issue #146, opened 2026-03-02
Release extension


jp-gouin:
The UI extension should be released into https://github.com/rancher/ui-plugin-charts  
Prime validation should be uncommented https://github.com/rancher/rancher-ai-ui/blob/d7d13b5877721624fb7e4bfab4114196d46dee1f/pkg/rancher-ai-ui/index.ts#L17 

Include the prime guardrail for the settings page

PR #157, opened 2026-03-09
Prime release

torchiaf:
Fixes https://github.com/rancher/rancher-ai-ui/issues/146

This change will enable Liz only to Prime users.
We also want to make the Chat available in Rancher 2.13, so we can't use `@shell/apis` to open the modal dialogs for compatibility reasons.
We want to move to the new API when all Rancher supported versions will expose the shell api. 
This issue is fro tracking it: https://github.com/rancher/rancher-ai-ui/issues/163

### DIFF
--- a/pkg/rancher-ai-ui/pages/settings/Settings.vue
+++ b/pkg/rancher-ai-ui/pages/settings/Settings.vue
@@ -5,7 +5,6 @@ import {
   onMounted,
 } from 'vue';
 import { useStore } from 'vuex';
-import { useShell } from '@shell/apis';
 import {
   AGENT_NAME, AGENT_NAMESPACE, AGENT_CONFIG_SECRET_NAME, AGENT_CONFIG_CONFIG_MAP_NAME, PERMISSIONS_DOCS_URL
 } from '../../product';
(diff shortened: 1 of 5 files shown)


pkg/rancher-ai-ui/pages/settings/Settings.vue
```diff
@@ -36,13 +36,18 @@ import { useChatApiComposable } from '../../composables/useChatApiComposable';
...
-const shellApi = useShell();
+// const shellApi = useShell();
 
 // TODO: All settings will be fetched through the API in the future.
 const { fetchSettings, saveSettings } = useChatApiComposable();
 
 const isLoading = ref(true);
+const isSaving = ref(false);
+
+const applySettingsBtn = ref<((arg: boolean) => void) | null>(null); // eslint-disable-line no-unused-vars
```
rak-phillip: AsyncButton exports a `AsyncButtonCallback` type that we should be able to import. If we do so, we should be able to drop the `no-unused-vars` specification and make the type a little easier to read: 

```suggestion
const applySettingsBtn = ref<AsyncButtonCallback | null>(null);
```

PR #170, opened 2026-03-13
Improve messages from hooks by adding the namespace

torchiaf:


pkg/rancher-ai-ui/handlers/hooks/template-message.ts
```diff
@@ -66,9 +67,10 @@ class TemplateMessageFactory {
         issue: issueText
       }, true);
       messageContent = t('ai.message.template.message.explainStateForResource', {
-        state: resource.state,
-        kind:  resource.kind,
-        name:  resource.name
+        state:     resource.state,
+        kind:      resource.kind,
+        name:      resource.name,
+        namespace: resource.namespace || null
```
marcelofukumoto: **Comment**
Same here. I wonder why is the only one that needs || null

PR #172, opened 2026-03-16
Fix sliding badges position in ResourceTables

### LINKED ISSUE #122: [Bug] Sliding badges break some tables
When the mouse goes hover the status badge in a table with the group by option on:

Image

torchiaf:
Fixes https://github.com/rancher/rancher-ai-ui/issues/122

The sliding badges are currently mounted as siblings of the other `td` elements in the Resource Tables. This is causing some glitches since the new sibling pushes the other columns to the right of the status column.

We are fixing the issue by mounting the badges as child of table's `tbody` element so that it doesn't collide with the table's columns.

### Screenshots

Before

https://github.com/user-attachments/assets/18b55ca4-ca38-4b12-96b4-10d26fddf30c

After

https://github.com/user-attachments/assets/b6147a3e-8bcb-481e-8416-5e659cde38e8

### DIFF
--- a/pkg/rancher-ai-ui/handlers/hooks/overlay/badge-sliding.ts
+++ b/pkg/rancher-ai-ui/handlers/hooks/overlay/badge-sliding.ts
@@ -174,7 +174,8 @@ class BadgeSlidingOverlay extends HooksOverlay {
     const badgeRect = badge.getBoundingClientRect();
     const badgeStyle = getComputedStyle(badge);
 
-    overlay.setAttribute('data-testid', 'rancher-ai-ui-sliding-badge');
+    overlay.setAttribute('data-testid', 'rancher-ai-ui-hook-overlay');
+
     overlay.classList.add(`${ HooksOverlay.defaultClassPrefix }-${ this.getSelector() }`);
     overlay.style.zIndex = '10';
     overlay.style.backgroundColor = overlayProps.background;
--- a/pkg/rancher-ai-ui/handlers/hooks/overlay/badge-sliding.ts
+++ b/pkg/rancher-ai-ui/handlers/hooks/overlay/badge-sliding.ts
@@ -2,7 +2,7 @@ import { nextTick } from 'vue';
 import { Store } from 'vuex';
 import { useI18n } from '@shell/composables/useI18n';
 import { warn } from '../../../utils/log';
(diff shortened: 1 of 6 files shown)


pkg/rancher-ai-ui/handlers/hooks/overlay/badge-sliding.ts
```diff
@@ -144,19 +146,21 @@ class BadgeSlidingOverlay extends HooksOverlay {
...
+        overlay.__containerPositionCleanup();
       }
     } catch (e) {
-      warn('Error during parent position cleanup', e);
+      warn('Error during container position cleanup', e);
     }
   }
 
   create(store: Store<any>, target: HTMLElement, badge: HTMLElement, ctx: Context, globalCtx: Context[] = []) {
+    this.hookContextTag = ctx.tag as HookContextTag;
```
marcelofukumoto, line 162: **Suggestion**
If you do like this you have only ONE context of the tag that you are going to use.
Is that enough?
From what I can see the CREATE maybe can create more than one, but you only record one context tag.

You can have 2 solutions.
Or you add a MAP that contains all the create so you can handle it separately.

It does feel you can have many overlays working at the same time. And if that does happen you have 2 or more creates but loses the reference of the first one.

Maybe I am going too deep, but feels like that when you just replace it here.

torchiaf: As soon as 1 hook is registered in the `HooksHandler.init()` , the `toggleOverlays` takes care of creating/destroying the elements when the users activate the overlays by mouse hover:

```
      if (show) {
        overlay.create(store, target, el, ctx, store.getters['rancher-ai-ui/context/all']);
      } else if (!(el.matches(':hover') || (el.querySelector(':hover') !== null))) {
        overlay.destroy(target);
      }
```

So - besides possible bugs - the create runs only 1 time for each hook depending on user's actions and we expect the destroy to be called after that (mouse out)

marcelofukumoto: If there is only one, should be ok. 
Ideally you could delete it if that is the case just to be sure it will be deleted.

marcelofukumoto (approved):
LGTM

Issue #141, opened 2026-02-24
The action links are always disabled for Rancher Projects


torchiaf:
The Chat is not able to fetch the Project's schema and than to enable the action button:

<img width="374" height="895" alt="Image" src="https://github.com/user-attachments/assets/cb2e1e00-a5ce-4ccf-9a66-6faac251576a" />

<img width="1612" height="186" alt="Image" src="https://github.com/user-attachments/assets/8ee5c6d5-8594-41b1-8135-a46e92f20934" />

deepakpunia-suse:
The node querying issue is now resolved, and the links are clickable. So closing this issue. 

<img width="641" height="346" alt="Image" src="https://github.com/user-attachments/assets/fc186c33-23f5-4616-b9e1-ebd3e6115e5f" />

PR #180, opened 2026-03-18
Add missing Resouce Ids conversion

torchiaf:
Fixes https://github.com/rancher/rancher-ai-ui/issues/141

- Fixing the NODE id - no conversion required
- Adding PodMetrics and NodeMetrics


marcelofukumoto (approved):
LGTM
Still has the schema problems. I would say adding some comments to explain the reasoning as a NOTE or TODO would be great. But it is good as it is.

PR #197, opened 2026-04-20
Handle links in the message bubbles to avoid opening them in the same…

torchiaf:
… window

### DIFF
--- a/pkg/rancher-ai-ui/components/message/index.vue
+++ b/pkg/rancher-ai-ui/components/message/index.vue
@@ -87,17 +87,6 @@ function handleShowThinking() {
   }));
 }
 
-// Handle click on links in the message content to avoid opening them in the same window
-function handleLinkClick(event: MouseEvent) {
-  const target = event.target as HTMLElement;
-  const link = target.closest('a');
-
-  if (link && link.href) {
-    event.preventDefault();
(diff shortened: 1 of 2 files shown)


pkg/rancher-ai-ui/components/message/index.vue
```diff
@@ -87,6 +87,17 @@ function handleShowThinking() {
   }));
 }
 
+// Handle click on links in the message content to avoid opening them in the same window
+function handleLinkClick(event: MouseEvent) {
```
rak-phillip: My only concern with this approach is that accessibility might suffer a bit with screen readers not being able to announce the behavior of the anchor tags. If I understand the scope of this PR correctly, it appears that we might be able to utilize markdown-it to create a custom renderer rule[^1] to add `target="_blank"` and `rel="noopener noreferrer"` to our links instead of intercepting. 

[^1]: https://github.com/markdown-it/markdown-it/blob/master/docs/examples/renderer_rules.md#reusing-existing-rules

torchiaf: Yes - this is a good idea, it's much cleaner now, thanks!

BTW the goal is to avoid opening any links generated by llm in the same window and losing Rancher.

rak-phillip (approved):
LGTM - Rerunning failed e2e tests to see if they pass.

PR #199, opened 2026-04-27
Move Chat scroll controlls to a new composable

torchiaf:


pkg/rancher-ai-ui/composables/useScrollComposable.ts
```diff
@@ -0,0 +1,90 @@
...
+      return;
+    }
+
+    nextTick(() => {
+      autoScrollEnabled.value = container.scrollTop + container.clientHeight >= container.scrollHeight - 2;
+      fastScrollEnabled.value = container.scrollTop + container.clientHeight < container.scrollHeight - 150;
+
+      // Save scroll position on every scroll event
+      lastScrollPosition.value = container.scrollTop;
+    });
```
marcelofukumoto: **Question**
Do we need really a nextTick? I was just wondering since handleScroll triggers it and should have the latest without the usage of the NextTick.
I tried with and without it and didnt see a difference. Maybe the use case for it to be required is something different than I tried.

And if it is required, maybe would be nice to add the reason.

pkg/rancher-ai-ui/composables/useScrollComposable.ts
```diff
@@ -0,0 +1,90 @@
+import {
+  ref, nextTick, watch, onMounted, onBeforeUnmount, Ref,
```
marcelofukumoto: **Nit**
Ref is a type, so you can add type Ref, or add in a new line.

marcelofukumoto (commented):
Overall works well! I didnt find any problem but added some comments there.
It is a great code improvement overall so thanks for that!

pkg/rancher-ai-ui/composables/useScrollComposable.ts
torchiaf: You are right, the next tick is not required here, so better removing it

marcelofukumoto (commented):
Just one thing I want to ask, let me just ask here. Usually for scrolls it feels better to debounce or throttle.
Throttle on this case.
It is not triggering every single scroll, I think that is usually not good, so a simple throttle even like a 100ms one would be interesting no?
What do you think?

marcelofukumoto (approved):
LGTM

PR #200, opened 2026-04-27
Fix console warnings

torchiaf:
This will fix some deprecations and remove various warnings from the console logs.

- Replace ghost,primary,secondary,tertiary modifiers with `variant={modifier}` syntax
- We are encapsulating some element in a parent div to avoid `extraneous class...` warnings

Note

- We are keeping `size` modifier as is, since `variant="{size}"` changes the style - we have to review the new dimension system with UX before to change it

### DIFF
--- a/pkg/rancher-ai-ui/components/context/SelectContext.vue
+++ b/pkg/rancher-ai-ui/components/context/SelectContext.vue
@@ -91,7 +91,7 @@ function reset() {
         @update:open="isOpen = $event"
       >
         <rc-dropdown-trigger
-          ghost
+          variant="ghost"
           small
           class="context-trigger"
           :disabled="props.disabled"
@@ -144,7 +144,7 @@ function reset() {
     >
       <RcButton
         small
-        tertiary
+        variant="tertiary"
         @click="reset"
       >
         <i class="icon icon-refresh mr-5" />
--- a/pkg/rancher-ai-ui/components/context/SelectContext.vue
+++ b/pkg/rancher-ai-ui/components/context/SelectContext.vue
@@ -85,50 +85,51 @@ function reset() {
     v-if="props.options.length > 0"
     class="context-select"
   >
-    <rc-dropdown
-      class="context-dropdown"
-      placement="top-end"
-      @update:open="isOpen = $event"
-    >
-      <rc-dropdown-trigger
-        ghost
-        small
-        class="context-trigger"
-        :disabled="props.disabled"
+    <div class="context-dropdown">
+      <rc-dropdown
+        placement="top-end"
+        @update:open="isOpen = $event"
       >
-        <span
-          class="context-trigger-text"
-          :class="{ 'ml-5': props.disabled }"
+        <rc-dropdown-trigger
+          ghost
+          small
+          class="context-trigger"
+          :disabled="props.disabled"
         >
-          {{ t('ai.context.add') }}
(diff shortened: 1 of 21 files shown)


pkg/rancher-ai-ui/components/context/SelectContext.vue
```diff
@@ -85,50 +85,51 @@ function reset() {
...
-      class="context-dropdown"
-      placement="top-end"
-      @update:open="isOpen = $event"
-    >
-      <rc-dropdown-trigger
-        ghost
-        small
-        class="context-trigger"
-        :disabled="props.disabled"
+    <div class="context-dropdown">
```
torchiaf, line 92: The only change in this block is: we are encapsulating everything in the `<div class="context-dropdown">` to avoid extraneous class warnings

pkg/rancher-ai-ui/pages/settings/Settings.vue
```diff
@@ -524,140 +524,146 @@ onMounted(async() => {
...
-  >
-    <h1 class="m-0">
-      {{ t('aiConfig.form.header') }}
-    </h1>
-
-    <Banner
-      v-if="!permissions?.list?.canListDeployments || !permissions?.list?.canListConfigMaps || !permissions?.list?.canListAiAgentCRDS"
-      class="m-0"
-      color="warning"
+  <div class="ai-configs-container">
```
torchiaf, line 527: The only change in this block is: we are encapsulating everything in the <div class="ai-configs-container"> to avoid extraneous class warnings

pkg/rancher-ai-ui/components/context/SelectContext.vue
marcelofukumoto, line 92: **Comment**
This context-dropdown is not styled right? Not sure what are the best practices that we follow. But if we are using just to "name" it, than it is ok.

pkg/rancher-ai-ui/components/context/SelectContext.vue
```diff
@@ -85,50 +85,51 @@ function reset() {
...
+            :key="i"
+            v-clean-tooltip="opt.description"
+            @click="toggleItem(opt)"
+          >
+            {{ opt.tag }}:{{ contextLabel(opt) }}
             <i
-              v-if="opt.icon"
-              class="icon"
-              :class="opt.icon"
+              v-if="selected.find((s: Context) => _id(s) === _id(opt))"
```
marcelofukumoto: **Nit**
From the pkg/rancher-ai-ui/components/agent/SelectAgent.vue you moves an inline .find to a computed variable. Would be nice to change it here as well. Just an opportunity :)

pkg/rancher-ai-ui/components/context/SelectContext.vue
torchiaf: It's not styled - it's used just to identify the container

pkg/rancher-ai-ui/components/context/SelectContext.vue
torchiaf: Done, thanks

marcelofukumoto: Oh! It even improve the class selection. Nice.

marcelofukumoto (approved):
LGTM

PR #205, opened 2026-05-05
Support Custom Headers secrets

### LINKED ISSUE #188: Support custom headers for the MCP authentication
MCP auth configuration should support a secret with custom headers.
This would allow the SUSE Observability MCP to work in Rancher. 
https://documentation.suse.com/cloudnative/suse-observability/latest/en/use/ai-assistant/mcp-server.html

torchiaf:
Fixes https://github.com/rancher/rancher-ai-ui/issues/188
Related PR https://github.com/rancher/rancher-ai-agent/pull/183

Liz now supports a new header authentication type for the Agents configs.
When set, the agent factory loads custom HTTP headers from the authenticationSecret and passes them to the MCP client connection.

We are adding the new `HEADER` type to support the authentication secret selection

- When the new type is selected, we can select a secret, we don't want inline creation for those secrets
- When switching from HEADER to BASIC and viceversa, we clean up the basic secrets queue and restore the original secret value

https://github.com/user-attachments/assets/8d171f50-e8e3-4973-aed4-353d0cb96c39

### DIFF
--- a/pkg/rancher-ai-ui/pages/settings/sections/__tests__/AIAgentConfigs.test.ts
+++ b/pkg/rancher-ai-ui/pages/settings/sections/__tests__/AIAgentConfigs.test.ts
@@ -1866,7 +1866,7 @@ describe('AIAgentConfigs.vue', () => {
         expect(vm.agentSecrets['test-agent']).toBeUndefined();
       });
 
-      it('should emit update:authentication-secrets event with agentSecrets when changing to non-BASIC auth', () => {
+      it('should emit undefined for update:authentication-secrets when changing away from BASIC auth', () => {
         const agent = mockAgent({
           spec: {
             ...mockAgent().spec,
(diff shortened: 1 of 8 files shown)


pkg/rancher-ai-ui/pages/settings/sections/__tests__/AIAgentConfigs.test.ts
```diff
@@ -1736,4 +1736,418 @@ describe('AIAgentConfigs.vue', () => {
...
+
+        const vm = wrapper.vm as any;
+
+        // No secrets stored
+        vm.updateAuthType(AIAgentConfigAuthType.HEADER);
+
+        const emitted = wrapper.emitted('update:authentication-secrets');
+
+        expect(emitted).toBeTruthy();
+        // Should emit undefined since we're not in BASIC mode
```
rak-phillip: The test title says "should not emit undefined", but the comment and the assertion say otherwise.

torchiaf: Fixed, thanks! Actually the test above this one is duplicate so I removed it as well

marcelofukumoto (changes requested):
Code LGTM but there are 2 bugs.
When you just create the data and save it it doesnt save your option.
So I believe after you SAVE you need to add the data to the cache and make sure it follows the same flow as when you reload the page.

Second bug is similar. I believe the cache is by position? If you reload the page, and delete the agent (MINUS button), you have 2 problems. **You cant apply anymore** and the data for the secret is cached.

**edit.: The apply works**

torchiaf:
> Code LGTM but there are 2 bugs. When you just create the data and save it it doesnt save your option. So I believe after you SAVE you need to add the data to the cache and make sure it follows the same flow as when you reload the page.
> 
> Second bug is similar. I believe the cache is by position? If you reload the page, and delete the agent (MINUS button), you have 2 problems. **You cant apply anymore** and the data for the secret is cached.
> 
> **edit.: The apply works**

Thanks @marcelofukumoto , I fixed both issues

marcelofukumoto (approved):
LGTM
It is working fine now.

PR #206, opened 2026-05-05
Support MCP's Oauth2

### LINKED ISSUE #186: Support OAuth2
The UI extension should support OAuth2. 
the Rancher MCP now supports OAuth2, this should be used instead of the Rancher Authentication. 

This would also allows for third-party Agent with MCP that uses OAuth2

torchiaf:
Fixes https://github.com/rancher/rancher-ai-ui/issues/186

This change introduces the possibility to setup the Oauth2 authentication when defining the agents and handle the auth requests in the Chat.

### Settings page

The user can setup the Oauth2 fields in the settings page:

image

image

We provide copy-paste URLs to setup the third-party Oauth applications:

image

### Chat panel 

The auth requests are sent to the Chat, when an agent which requires oauth2 verification is triggered. This happens for each agent executed in the single message.

https://github.com/user-attachments/assets/42ad4d46-d2c1-4504-9f8e-0e947aca5869

Before initializing new chats, the server needs to check the MCP connection of each agent in the system.
When there are many agents in the system, this can take several seconds (even ~10 seconds depending on MCP servers).

Until the chat is not ready, we want to show the new  "Preparing chat" label to be displayed in between the ws connection is open and the chat metadata is received from the server.
We are also disabling the context and the console panel to avoid any interaction with the chat while the transition is on going.

image

### Key Changes

See comments below

### DIFF
--- a/pkg/rancher-ai-ui/components/DiscoveryBanner/DiscoveryBanner.vue
+++ b/pkg/rancher-ai-ui/components/DiscoveryBanner/DiscoveryBanner.vue
@@ -8,6 +8,15 @@ interface DiscoveryStatus {
   message?: string;
 }
 
+type BannerAction = 'confirm' | 'cancel';
+
+interface BannerProps {
+  color: string;
+  key: string;
+  message: string;
+  action: BannerAction;
+}
+
 const props = defineProps({
   status: {
     type:    Object as () => DiscoveryStatus,
@@ -21,7 +30,7 @@ const props = defineProps({
 
 const emit = defineEmits(['confirm', 'cancel']);
 
-const banner = computed(() => {
+const banner = computed<BannerProps>(() => {
   const status = props.status;
 
   const color = status?.result || 'info';
@@ -36,6 +45,10 @@ const banner = computed(() => {
     action,
   };
 });
+
+function emitAction(value: BannerAction) {
+  emit(value);
+}
 </script>
 
 <template>
@@ -57,7 +70,7 @@ const banner = computed(() => {
           v-clean-html="content"
           class="clickable-label"
           :data-testid="`rancher-ai-ui-discovery-banner-button-${props.translationKey}`"
-          @click="emit(banner.action as any)"
+          @click="emitAction(banner.action)"
         />
       </template>
     </RichTranslation>
--- a/pkg/rancher-ai-ui/components/DiscoveryBanner/DiscoveryBanner.vue
+++ b/pkg/rancher-ai-ui/components/DiscoveryBanner/DiscoveryBanner.vue
@@ -1,0 +1,72 @@
+<script setup lang="ts">
+import { computed } from 'vue';
+import Banner from '@components/Banner/Banner.vue';
+import RichTranslation from '@shell/components/RichTranslation.vue';
+
+interface DiscoveryStatus {
+  result?: 'success' | 'warning' | 'info' | null;
+  message?: string;
+}
+
+const props = defineProps({
+  status: {
+    type:    Object as () => DiscoveryStatus,
+    default: () => ({} as DiscoveryStatus),
+  },
+  translationKey: {
+    type:    String,
+    default: '',
+  },
+});
+
+const emit = defineEmits(['confirm', 'cancel']);
+
+const banner = computed(() => {
+  const status = props.status;
+
+  const color = status?.result || 'info';
+  const key = status?.result === undefined ? 'info' : status?.result || 'inProgress';
+  const message = status?.message || '';
+  const action = status?.result === null ? 'cancel' : 'confirm';
+
+  return {
+    color,
+    key,
+    message,
+    action,
+  };
+});
+</script>
+
+<template>
+  <Banner
+    class="m-0"
+    :color="banner.color"
+  >
+    <RichTranslation
+      :k="`${props.translationKey}.${banner.key}`"
+    >
+      <template #message>
+        <span
+          v-if="banner.message"
+          v-clean-html="t(`${props.translationKey}.message`, { message: banner.message }, true)"
+        />
+      </template>
+      <template #action="{ content }">
+        <span
+          v-clean-html="content"
+          class="clickable-label"
+          :data-testid="`rancher-ai-ui-discovery-banner-button-${props.translationKey}`"
+          @click="emit(banner.action as any)"
+        />
+      </template>
+    </RichTranslation>
+  </Banner>
+</template>
+
+<style scoped lang="scss">
+.clickable-label {
+  text-decoration: underline;
+  cursor: pointer;
+}
+</style>
--- a/pkg/rancher-ai-ui/components/message/template/McpAuthenticationRequest.vue
+++ b/pkg/rancher-ai-ui/components/message/template/McpAuthenticationRequest.vue
@@ -5,6 +5,16 @@ import { ConfirmationStatus, Message } from '../../../types';
 import { formatMessageContent } from '../../../utils/format';
 import SystemAvatar from '../avatar/SystemAvatar.vue';
 
+interface Agent {
+  displayName: string;
+  description: string;
+}
+
+interface Content {
+  message?: string;
+  agent?: Agent;
+}
+
 const props = defineProps({
   message: {
     type:    Object as PropType<Message>,
@@ -17,6 +27,8 @@ const props = defineProps({
 });
 
 const emit = defineEmits(['update:message']);
+
+const content = computed<Content | undefined>(() => props.message.templateContent?.content);
 
 const actions = computed(() => {
   const all = props.message?.actions || [];
@@ -46,25 +58,25 @@ function doAction(type: 'confirm' | 'cancel') {
   >
     <SystemAvatar class="chat-msg-avatar" />
     <div
-      v-if="props.message?.templateContent?.content?.message"
+      v-if="content?.message"
       class="chat-system-mcp-request-request-msg-bubble"
     >
       <div class="chat-system-mcp-request-request-msg-text">
         <div
-          v-if="props.message?.templateContent?.content?.agent"
+          v-if="content?.agent"
           data-testid="rancher-ai-ui-chat-message-selected-agent-label"
           class="chat-system-mcp-request-request-msg-agent"
         >
           <span
-            v-clean-tooltip="props.message.templateContent.content.agent.description"
+            v-clean-tooltip="content.agent.description"
             class="chat-system-mcp-request-request-msg-agent-label"
           >
-            {{ t('ai.agents.selectedAgent.label', { agent: props.message.templateContent.content.agent.displayName }) }}
+            {{ t('ai.agents.selectedAgent.label', { agent: content.agent.displayName }) }}
           </span>
         </div>
 
         <span
-          v-clean-html="formatMessageContent(props.message.templateContent.content.message || '')"
+          v-clean-html="formatMessageContent(content.message || '')"
         />
       </div>
 
@@ -182,8 +194,10 @@ function doAction(type: 'confirm' | 'cancel') {
   }
 }
 
+@import '@shell/assets/styles/base/_color-classic.scss';
+
 .theme-dark .chat-system-mcp-request-request-msg-text :deep(code) {
-  color: #C0EFDE;
+  color: $green-40;
 }
 
 .chat-system-mcp-request-request-actions {
--- a/pkg/rancher-ai-ui/components/message/template/McpAuthenticationRequest.vue
+++ b/pkg/rancher-ai-ui/components/message/template/McpAuthenticationRequest.vue
@@ -117,7 +117,7 @@ function doAction(type: 'confirm' | 'cancel') {
 <style lang='scss' scoped>
 .chat-system-mcp-request-request-message {
   display: flex;
-  gap: 8px;
+  gap: var(--gap);
 }
 
 .chat-system-mcp-request-request-msg-bubble {
@@ -189,12 +189,12 @@ function doAction(type: 'confirm' | 'cancel') {
 .chat-system-mcp-request-request-actions {
   display: flex;
   justify-content: flex-end;
-  gap: 8px;
+  gap: var(--gap);
   margin-top: 16px;
 
   &-buttons {
     display: flex;
-    gap: 8px;
+    gap: var(--gap);
   }
 }
 
--- a/pkg/rancher-ai-ui/components/message/template/McpAuthenticationRequest.vue
+++ b/pkg/rancher-ai-ui/components/message/template/McpAuthenticationRequest.vue
@@ -1,0 +1,221 @@
+<script setup lang="ts">
+import { computed, type PropType } from 'vue';
+import RcButton from '@components/RcButton/RcButton.vue';
+import { ConfirmationStatus, Message } from '../../../types';
+import { formatMessageContent } from '../../../utils/format';
+import SystemAvatar from '../avatar/SystemAvatar.vue';
+
+const props = defineProps({
+  message: {
+    type:    Object as PropType<Message>,
+    default: () => ({} as Message),
+  },
+  disabled: {
+    type:    Boolean,
+    default: false,
+  },
+});
+
+const emit = defineEmits(['update:message']);
+
+const actions = computed(() => {
+  const all = props.message?.actions || [];
+
+  return {
+    confirm: all[0],
+    cancel:  all[1],
+  };
+});
+
+function doAction(type: 'confirm' | 'cancel') {
+  if (actions.value[type]?.action) {
+    const confirmation = actions.value[type].action();
+
+    emit('update:message', {
+      ...props.message,
+      confirmation
+    });
+  }
+}
+</script>
+
+<template>
+  <div
+    class="chat-system-mcp-request-request-message"
+    :class="{ 'disabled-panel': props.disabled }"
+  >
+    <SystemAvatar class="chat-msg-avatar" />
+    <div
+      v-if="props.message?.templateContent?.content?.message"
+      class="chat-system-mcp-request-request-msg-bubble"
+    >
+      <div class="chat-system-mcp-request-request-msg-text">
+        <div
+          v-if="props.message?.templateContent?.content?.agent"
+          data-testid="rancher-ai-ui-chat-message-selected-agent-label"
+          class="chat-system-mcp-request-request-msg-agent"
+        >
+          <span
+            v-clean-tooltip="props.message.templateContent.content.agent.description"
+            class="chat-system-mcp-request-request-msg-agent-label"
+          >
+            {{ t('ai.agents.selectedAgent.label', { agent: props.message.templateContent.content.agent.displayName }) }}
+          </span>
+        </div>
+
+        <span
+          v-clean-html="formatMessageContent(props.message.templateContent.content.message || '')"
+        />
+      </div>
+
+      <div class="chat-system-mcp-request-request-actions">
+        <div
+          v-if="props.message.confirmation"
+          :data-testid="`rancher-ai-ui-chat-message-confirmation-${ props.message.confirmation?.status === ConfirmationStatus.Confirmed ? 'confirmed' : 'canceled'}`"
+          :class="['chat-system-mcp-request-request-actions-result', `status-${props.message.confirmation.status}` ]"
+        >
+          <i
+            :class="[ 'icon', props.message.confirmation.icon ]"
+          />
+          <span
+            v-clean-html="formatMessageContent(props.message.confirmation.label || '')"
+            class="chat-system-mcp-request-request-msg-text"
+          />
+        </div>
+        <div
+          v-else
+          class="chat-system-mcp-request-request-actions-buttons"
+        >
+          <RcButton
+            v-if="actions.cancel?.action"
+            small
+            tertiary
+            data-testid="rancher-ai-ui-chat-message-confirmation-cancel-button"
+            @click="doAction('cancel')"
+          >
+            <span class="rc-button-label">
+              {{ actions.cancel?.label }}
+            </span>
+          </RcButton>
+          <RcButton
+            v-if="actions.confirm?.action"
+            small
+            tertiary
+            data-testid="rancher-ai-ui-chat-message-confirmation-confirm-button"
+            @click="doAction('confirm')"
+          >
+            <span class="rc-button-label">
+              {{ actions.confirm?.label }}
+            </span>
+          </RcButton>
+        </div>
+      </div>
+    </div>
+  </div>
+</template>
+
+<style lang='scss' scoped>
+.chat-system-mcp-request-request-message {
+  display: flex;
+  gap: 8px;
+}
+
+.chat-system-mcp-request-request-msg-bubble {
+  position: relative;
+  background: var(--body-bg);
+  color: var(--body-text);
+  border: 1px solid var(--border);
+  border-radius: 12px;
+  box-shadow: 0 2px 8px 0 rgba(0,0,0,0.04);
+  padding: 12px;
+  line-height: 21px;
+  display: flex;
+  flex-direction: column;
+  gap: 6px;
+
+  /* Hide actions by default, show on hover */
+  &:hover .chat-system-mcp-request-request-msg-bubble-actions {
+    opacity: 1;
+    pointer-events: auto;
+  }
+}
+
+.chat-system-mcp-request-request-msg-agent {
+  margin-bottom: 4px;
+}
+.chat-system-mcp-request-request-msg-agent-label {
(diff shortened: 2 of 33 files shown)


pkg/rancher-ai-ui/composables/useAIAgentApiComposable.ts
```diff
@@ -234,6 +375,11 @@ export function useAIAgentApiComposable(agents?: ComputedRef<Agent[]>) {
   }
 
   return {
+    refreshMcpAuthenticationToken,
```
torchiaf, line 402: The new Endpoints to handle the `metadata` and `clientInfo` discovery

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
```diff
@@ -296,6 +307,63 @@ export function useChatMessageComposable(
     }
   }
 
+  async function ensureAgentMcpAuthenticationRequest(ws: WebSocket, agents: Agent[], metadata: McpAuthenticationRequest) {
```
torchiaf, line 311: This is responsible for showing the Auth request message using `AuthenticationHandler`

pkg/rancher-ai-ui/handlers/authentication.ts
```diff
@@ -0,0 +1,113 @@
+import { markRaw } from 'vue';
```
torchiaf, line 1: We handle the auth popup and the communication between the main UI and the popup via `postMessage` here.

pkg/rancher-ai-ui/pages/settings/sections/ai-agent-configs/index.vue
```diff
@@ -473,37 +497,50 @@ watch(validationErrors, (errors) => {
...
+                <div class="col span-6">
+                  <SecretSelector
+                    :value="selectedAgent.spec.authenticationSecret || undefined"
+                    :secret-name-label="t('aiConfig.form.section.aiAgent.fields.authenticationSecret.label')"
+                    :namespace="AGENT_NAMESPACE"
+                    @update:value="(value: string) => updateAgent({ spec: { ...selectedAgent.spec, authenticationSecret: value || undefined } })"
+                  />
+                </div>
               </div>
+              <Oauth2SecretForm
```
torchiaf: `Oauth2SecretForm` is the component responsible for fetching the Oauth2 secrets, add the new value and handle the discovery info

pkg/rancher-ai-ui/pages/settings/Settings.vue
```diff
@@ -341,40 +346,127 @@ async function saveAiAgentConfigAuthenticationSecrets() {
...
+          },
+          data,
+        });
+
+        secret._type = SECRET_TYPES.BASIC;
+      }
 
-    secret._type = SECRET_TYPES.BASIC;
+      break;
+    case AIAgentConfigAuthType.OAUTH2:
```
torchiaf, line 379: We are now handling new Oauth2 secrets

pkg/rancher-ai-ui/pages/Chat.vue
```diff
@@ -425,15 +425,15 @@ function unmount() {
       />
       <Context
         :value="!hasPermissions || systemErrors.length ? [] : context"
-        :disabled="disabled"
+        :disabled="!isChatInitialized || disabled"
```
torchiaf, line 449: We want to disable the context and console panels when the chat is not initialized - this mitigate any delays when initializing new chats

pkg/rancher-ai-ui/store/chat.ts
```diff
@@ -48,6 +48,10 @@ const getters = {
     return state.chats[chatId]?.messages[messageId] || null;
   },
   processingState: (state: State) => (chatId: string) => {
+    if (!state.metadata?.chatId) {
+      return { phase: MessagePhase.Setup };
+    }
```
torchiaf, line 53: We want to show a "setup" message from when the websocket is connected to when the chat receives the `chat-metadata` - this can take several seconds depending on the number of agents in the system.

pkg/rancher-ai-ui/components/DiscoveryBanner/DiscoveryBanner.vue
```diff
@@ -0,0 +1,72 @@
...
+      </template>
+    </RichTranslation>
+  </Banner>
+</template>
+
+<style scoped lang="scss">
+.clickable-label {
+  text-decoration: underline;
+  cursor: pointer;
+}
```
marcelofukumoto, line 84: **Comment**
This feels like a normal link, why we need a specific class for this one?

pkg/rancher-ai-ui/components/message/template/McpAuthenticationRequest.vue
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  &:deep(th) {
+    text-align: left;
+  }
+
+  &:deep(pre) {
+    margin: 0;
+  }
+}
```
marcelofukumoto, line 195: Strong use of deep here. Isnt there a different alternative?
Feels a little forced.

pkg/rancher-ai-ui/components/message/template/McpAuthenticationRequest.vue
```diff
@@ -0,0 +1,221 @@
...
+
+.theme-dark .chat-system-mcp-request-request-msg-text :deep(code) {
+  color: #C0EFDE;
```
marcelofukumoto: That is green-40, so I think we can use the variable.

pkg/rancher-ai-ui/components/message/template/McpAuthenticationRequest.vue
```diff
@@ -0,0 +1,221 @@
...
+}
+
+.chat-system-mcp-request-request-actions {
+  display: flex;
+  justify-content: flex-end;
+  gap: 8px;
```
marcelofukumoto: A gap of 8px can always use the --gap instead

pkg/rancher-ai-ui/composables/useAIAgentApiComposable.ts
```diff
@@ -2,7 +2,7 @@ import { ComputedRef } from 'vue';
 import { AGENT_NAME, AGENT_NAMESPACE, AGENT_REST_API_PATH } from '../product';
 import { error } from '../utils/log';
 import {
-  Agent, AgentSettings, Context, HistoryChat, HistoryChatMessage, LLMProvider, Message,
+  Agent, AgentSettings, AiAgentAPIEvent, Context, HistoryChat, HistoryChatMessage, LLMProvider, Message,
```
marcelofukumoto: This is really Nit, but you used 
useAI with AI as capital
And here it is AiAgent, and the Ai is not capital
Really really picky! Sorry about it.

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
```diff
@@ -65,19 +74,21 @@ export function useChatMessageComposable(
...
-  const messageBox = computed(() => store.getters['rancher-ai-ui/chat/messageBox'](chatId));
-  const messages = computed(() => Object.values(store.getters['rancher-ai-ui/chat/messages'](chatId)) as Message[]);
   const currentMsg = ref<Message>({} as Message);
-  const error = computed(() => store.getters['rancher-ai-ui/chat/error'](chatId));
 
   const chatMetadata = computed<ChatMetadata>(() => store.getters['rancher-ai-ui/chat/metadata'] || {});
   const isChatInitialized = computed(() => !!chatMetadata.value.chatId);
 
-  const processingState = computed(() => store.getters['rancher-ai-ui/chat/processingState'](chatId) || { phase: MessagePhase.Initializing });
+  const messageBox = computed(() => store.getters['rancher-ai-ui/chat/messageBox'](chatId));
```
marcelofukumoto, line 87: Can we turn those `rancher-ai-ui/chat/messageBox` into const instead?
It feels we even have a prefix, so I would think it would organize a little or the mistake of writing down wrongly.

pkg/rancher-ai-ui/handlers/authentication.ts
```diff
@@ -0,0 +1,132 @@
...
+  private isCompleted: boolean = false;
+
+  private resolveFn: () => void;
+  private cancelFn: () => void;
+
+  constructor(metadata: McpAuthenticationRequest, resolveFn: () => void, cancelFn: () => void) {
+    this.resolveFn = resolveFn;
+    this.cancelFn = cancelFn;
+
+    const authWindow = window.open(metadata.url, `auth_${ metadata.type }`, 'width=600,height=700,left=200,top=200');
```
marcelofukumoto, line 24: It is better to sanitize the metadata.url in some way. It does feel that just opening a new window in a random URL can contain security risk.

pkg/rancher-ai-ui/types.ts
```diff
@@ -196,15 +205,12 @@ export const enum MessageTemplateComponent {
...
 }
 
 export interface MessageTemplate {
   component: MessageTemplateComponent;
-  content: {
-    message?: string;
-    principal?: any;
-    [key: string]: unknown;
-  };
+  content: { [key: string]: any; };
```
marcelofukumoto: Why changed from the proper content to just any? As much we have "known" interfaces would be better

pkg/rancher-ai-ui/components/DiscoveryBanner/DiscoveryBanner.vue
```diff
@@ -0,0 +1,72 @@
...
+          v-if="banner.message"
+          v-clean-html="t(`${props.translationKey}.message`, { message: banner.message }, true)"
+        />
+      </template>
+      <template #action="{ content }">
+        <span
+          v-clean-html="content"
+          class="clickable-label"
+          :data-testid="`rancher-ai-ui-discovery-banner-button-${props.translationKey}`"
+          @click="emit(banner.action as any)"
```
marcelofukumoto: The precise cast would be the banner actions. Which would be 'cancel' or 'confirm'
I don't really like this inline emit, to create a function would feel better for me, but it is ok. Only personal opinion.

pkg/rancher-ai-ui/handlers/authentication.ts
```diff
@@ -0,0 +1,132 @@
...
+const OAUTH2_CHANNEL_NAME = 'oauth_channel';
+const OAUTH2_SUCCESS_MESSAGE = 'oauth_success';
+
+/**
+ * Handles OAuth2 authentication requests.
+ */
+class OAuth2AuthenticationRequest {
+  private window: Window | null;
+  private broadcastChannel: BroadcastChannel | null = null;
+  private clearCloseInterval: any;
```
marcelofukumoto: Marking some `any`

pkg/rancher-ai-ui/handlers/authentication.ts
```diff
@@ -0,0 +1,132 @@
...
+  createOauth2AuthenticationRequest(
+    metadata: McpAuthenticationRequest,
+    resolveFn: () => void,
+    cancelFn: () => void
+  ): OAuth2AuthenticationRequest {
+    if (this.activeOauth2Request) {
+      this.activeOauth2Request.abort({ notify: false });
+    }
+
+    this.activeOauth2Request = markRaw(new OAuth2AuthenticationRequest(metadata, resolveFn, cancelFn));
```
marcelofukumoto, line 116: **Question**
Why use markRaw here is required?

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
```diff
@@ -296,6 +307,63 @@ export function useChatMessageComposable(
     }
   }
 
+  async function ensureAgentMcpAuthenticationRequest(ws: WebSocket, agents: Agent[], metadata: McpAuthenticationRequest) {
+    const agent = agents.find((a) => a.name === metadata?.agent);
+
+    if (agent) {
```
marcelofukumoto, line 314: **Question**
What happens if there is no agent?

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
```diff
@@ -333,7 +402,7 @@ export function useChatMessageComposable(
       }
     } else {
       try {
-        await processMessageData(data);
+        await processMessageData(ws, data);
       } catch (err) {
         processMessageErrorData(err as Error);
```
marcelofukumoto: I saw at the top that this should be ChatError no?

pkg/rancher-ai-ui/pages/Chat.vue
```diff
@@ -146,6 +147,24 @@ const systemErrors = computed(() => {
   }
 });
 
+const processingMessageState = computed(() => {
+  if (!hasPermissions) {
```
marcelofukumoto: hasPermission is always true since it is an object.
You need to use hasPermission.value

marcelofukumoto (commented):
Just the first check, I am running the app as well when taking a break of reading code.
For now that is what I have.

It feels quite complex, makes me wonder if we should have an architectural file of what is happening. Just a random input, that works for most tasks.

I will test a little more, but seems to be working fine.
If I find any problem I can try to check in the code the reason.

pkg/rancher-ai-ui/components/DiscoveryBanner/DiscoveryBanner.vue
torchiaf: Because normal link would look like:

Screenshot from 2026-07-08 17-24-57

image


Cursor and color would be wrong.

We want the color always grey and cursor pointer:

image

pkg/rancher-ai-ui/components/message/template/McpAuthenticationRequest.vue
torchiaf: Yeah - it's a bit wrong but we have this CSS rules in other components as well - I would move this refactoring to another PR and fix it everywhere in the code

pkg/rancher-ai-ui/composables/useAIAgentApiComposable.ts
torchiaf: np - it's an easy one

pkg/rancher-ai-ui/handlers/authentication.ts
torchiaf: Yeah, I agree, btw this should be done backend side.
Done

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
torchiaf: The store access in the UI is always by string - I agree about converting them into constants, but for now, I would leave things as they are and postpone this refactoring for the entire interface until later.

pkg/rancher-ai-ui/types.ts
torchiaf: Reverted

pkg/rancher-ai-ui/components/DiscoveryBanner/DiscoveryBanner.vue
torchiaf: Good catch, I left the `any` type in the code.
I cleaned up a little bit the component adding new types.

pkg/rancher-ai-ui/handlers/authentication.ts
torchiaf: Fixed

pkg/rancher-ai-ui/handlers/authentication.ts
torchiaf: It's for preventing Vue's reactivity.
I think is no more required since it doesn't access to the store - maybe it did in previous changes, but it's good to keep it to avoid errors on future code refactorings and for performance.

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
torchiaf: The AI agent's Helm chart always install 3 built-in agents. If all of them are `disabled`, the chat won't even start:

image

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
torchiaf: Fixed

pkg/rancher-ai-ui/pages/Chat.vue
torchiaf: Actually is a boolean, right ?

```
  const hasPermissions = computed(() => {
    const canListDeployments = !!store.getters['management/canList'](WORKLOAD_TYPES.DEPLOYMENT);
    const canListConfigMaps = !!store.getters['management/canList'](CONFIG_MAP);

    return canListDeployments && canListConfigMaps;
  });
  
```

torchiaf:
@marcelofukumoto thanks for the review. I addressed your comments and pushed 2 new enhancements:

- https://github.com/rancher/rancher-ai-ui/pull/206/changes/b2433bdc594c70271b59dded756f6455211d35cf - this comes after UX review, it's just a llok&fill refactoring - no new logics.
- https://github.com/rancher/rancher-ai-ui/pull/206/changes/9ef877662bca9c22e2c915873d5bef3870906aa8 - we are adding the copy-paste fields for the third-party Apps URLs.
- https://github.com/rancher/rancher-ai-ui/pull/206/changes/0945d757a67e744f58932c6b3e17ed85e333bfd4 - fixed Apps URLs after UX review.

The other fixes are from your review.
Please take a look.

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
```diff
@@ -475,6 +549,40 @@ export function useChatMessageComposable(
...
+        const metadata = formatMcpAuthenticationRequest(data);
+
+        if (!metadata || !validateUrl(metadata.url)) {
+          throw {
+            message: t('ai.error.oauth2.invalidMetadata'),
+            key:     'message'
+          };
+        }
+
+        await ensureAgentMcpAuthenticationRequest(ws, agents.value, metadata);
```
marcelofukumoto, line 562: **Question**
We are waiting for the message to follow up.
If there are nothing it will just keep waiting. Should be ok since as you said that maybe not a real case.
But if happens than it will be stuck. Would be nice to have a flow here as well. Like "automatically" sending a Cancel or anything.

What I am saying is: If renders, perfect. If doesn't render. Than it is stuck because it will never receive the response.

pkg/rancher-ai-ui/handlers/authentication.ts
```diff
@@ -0,0 +1,136 @@
...
+    }, 500);
+
+    this.window = authWindow;
+  }
+
+  public abort(args = { notify: true }): void {
+    if (this.clearCloseInterval) {
+      clearInterval(this.clearCloseInterval);
+
+      if (args?.notify) {
```
marcelofukumoto, line 72: **Question/Nit**
Seeing the difference between this and the cancel on the top. They are similar, but one is aborting and the other one is really to close the interval.

I was just wondering if we can check if `!this.isCompleted` here just to make sure the cancel is not send after it is already complete. It seems like an edge case so maybe a Nit, but could happen specially if the request is almost done.

pkg/rancher-ai-ui/pages/settings/Settings.vue
```diff
@@ -341,40 +346,127 @@ async function saveAiAgentConfigAuthenticationSecrets() {
...
+        const secret = await store.dispatch(`management/find`, {
+          type: SECRET,
+          id:   `${ AGENT_NAMESPACE }/${ secretName }`,
+          opt:  { watch: true }
+        });
+
+        if (secret) {
+          await secret.remove();
+        }
+      } catch (error) {
```
marcelofukumoto: **Nit**
We have `catch (error)`, `catch (err)` and `catch (e)`

pkg/rancher-ai-ui/utils/format.ts
```diff
@@ -131,29 +132,48 @@ export function formatAgentMetadata(data: string): Partial<AgentMetadata> | null
...
 
+export function formatMcpAuthenticationRequest(data: string): McpAuthenticationRequest | null {
+  const cleaned = data.replaceAll(Tag.AuthenticationRequestStart, '').replaceAll(Tag.AuthenticationRequestEnd, '').trim();
+
+  try {
+    const parsed = JSON.parse(cleaned);
+
+    return parsed;
+  } catch (error) {
+    console.error('Failed to parse MCP authentication request:', error); /* eslint-disable-line no-console */
```
marcelofukumoto: **Question**
Do we need to use the console? Can't be an error like the others?

pkg/rancher-ai-ui/utils/url.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    return false;
+  }
+
+  // Block potentially dangerous protocols
+  if (value.includes('javascript:') || value.includes('data:')) {
+    warn('URL contains invalid protocol');
+
+    return false;
+  }
```
marcelofukumoto: **Question**
Nice to have additional protection. But we check the protocol just after, so that is just additional? Is it required?

pkg/rancher-ai-ui/pages/Chat.vue
marcelofukumoto: image

I debugged on my side, I think it is returning the object on that level. 
The hasPermissions.value that has the boolean.

Or I am doing something wrong?

marcelofukumoto (commented):
I checked, still working. I am not even sure what changed. But I setup everything over again and seems to be working fine.

Just added some points since there was changes.

There are 2 unresolved points from the previous one.
But I think we are almost done.

There is one blocker that is the IF there that I am pretty sure it is an object. But overall, all good. Next one should be the last.

pkg/rancher-ai-ui/components/message/template/McpAuthenticationRequest.vue
torchiaf: Sorry I missed this one, fixed

pkg/rancher-ai-ui/pages/Chat.vue
torchiaf: Ok sorry, I misread your initial comment. it's a `ComputedRef` so the raw value is boolean but here is checking the ref.
Fixed:

```
  if (!hasPermissions.value) {
    return null;
  }
```

pkg/rancher-ai-ui/handlers/authentication.ts
torchiaf: We don't want to check `isCompleted` here because we want to abort the request without sending the `cancel` event, explicitly.
There is only one case where `notify` is false:

```
  createOauth2AuthenticationRequest(
    metadata: McpAuthenticationRequest,
    resolveFn: () => void,
    cancelFn: () => void
  ): OAuth2AuthenticationRequest {
    if (this.activeOauth2Request) {
      this.activeOauth2Request.abort({ notify: false }); // Here
    }

    this.activeOauth2Request = markRaw(new OAuth2AuthenticationRequest(metadata, resolveFn, cancelFn));

    return this.activeOauth2Request;
  }
```

This is an edge case - it should never happen, but if this is the case, it means we ended up in a situation where there is 1 McpAuth message request pending and a new one is created in the UI.
In this scenario, it means that the server is already awaiting for the event to resolve the new auth request -> so we want to destroy the previous request without sending the cancel event, which otherwise would be considered as an event for the new request  by the server, which would be wrong.

pkg/rancher-ai-ui/pages/settings/Settings.vue
torchiaf: Fixed

pkg/rancher-ai-ui/utils/format.ts
torchiaf: Oh yes, fixed

pkg/rancher-ai-ui/utils/url.ts
torchiaf: Agree, this is not required, removed

pkg/rancher-ai-ui/handlers/authentication.ts
marcelofukumoto: All good! I believe it is an edge case. It should not happen, I am not sure on the response time after the confirmation. So it could be more prominent than it seems.
But all good, even if happens doesn't feel like it is a big impact since just authentication.

marcelofukumoto (approved):
LGTM

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
torchiaf: The await is only for the `addMessage`, to keep the messag count consistent. The rest of the auth workflow happens via ws handshake

PR #211, opened 2026-05-12
UI tools Preview

torchiaf:
image


oboc-sts (changes requested):
As per the recording I've sent over Slack, we need to do a few changes before merging.

torchiaf:
Thanks @oboc-sts this the layout's last look, based on your review - let's refine it if needed, otherwise we can go with the technical review.


https://github.com/user-attachments/assets/43ba65db-ee15-465f-8dc5-f2a65d607979

oboc-sts (approved):
Looking sharp after the latest changes. It's ready to be merged from my pov 🚀

marcelofukumoto:
@torchiaf I just executed my auto-review and got many details.
Maybe you want to check before I go full throttle on the code and dev test:
https://code-review-share.13.48.147.135.sslip.io/r/urvKABozso_EZDiB

torchiaf:
Thanks @marcelofukumoto , given trhe suggestions from the auto-review:

image


- I pushed some changes, mostly to improve the accessibility.
- I also updated the image's base path:

```
const BASE_PATH = 'https://raw.githubusercontent.com/rancher/rancher-ai-ui/main/assets/ui-tools/screenshots';
```

In order to test the feature in dev, replace it with the link to the current PR's branch:
```
const BASE_PATH = 'https://raw.githubusercontent.com/torchiaf/rancher-ai-ui/feature-ui-tools-preview/assets/ui-tools/screenshots';
```

- IMO  the other comments in the review are negligible.

marcelofukumoto:
Perfect! Let me rerun to get a feel of it and start checking manually!

marcelofukumoto:
i got a nice rerun
https://code-review-share.13.48.147.135.sslip.io/r/qn28Xo8WUYPCbSbZ
Still getting one blocker which I agree, seems like a nice change to have.

The other suggestions I would focus on the:
- Focus trap one
- The loading one does make sense for a better UX as well
- The en-us clean up
- There are some interesting clean ups as well of unused things

So not many changes, mostly are e2e test ones.

marcelofukumoto (commented):
i got a nice rerun
https://code-review-share.13.48.147.135.sslip.io/r/qn28Xo8WUYPCbSbZ
Still getting one blocker which I agree, seems like a nice change to have.

The other suggestions I would focus on the:

Focus trap one
The loading one does make sense for a better UX as well
The en-us clean up
There are some interesting clean ups as well of unused things
So not many changes, mostly are e2e test ones.

torchiaf:
Thanks @marcelofukumoto I pushed the fixes, but i didn't find anything relevant related to the e2e. Consider that the new e2e tests producing the snapshots are launched by developers manually, we don't to worry too much about the `cy.wait` and cleanups

marcelofukumoto (commented):
I am doing some tests. 
1. Clicking outside the panel is not supposed to close it?
2. The images are broken for me right? So I guess that is expected for now.
3. Something seems off with the loading when I click, it shows in a weird way. Maybe you can check in slow motion to see what is doing.

torchiaf:
1. Right, I pushed a new commit to close the panel when clicking outside.
2. They are broken because the link is pointing to the images in the upstream repo - this will work as soon as we push the PR
3. What issue do you see ? I have this:
    - Normal loading (Broken images, see previous point)

    https://github.com/user-attachments/assets/79f91523-a98c-4db8-81b8-ad7c419b566b

    - Mock loading
  
    https://github.com/user-attachments/assets/a04684a2-0d81-494a-a15e-e7d3835c0c89

marcelofukumoto:
Ok! thanks!
For the loading it was just because it was quite fast so it felt weird but I didnt investigate more. Seems to be working fine

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/index.vue
```diff
@@ -472,6 +485,9 @@ const resetToolsToDefaults = () => {
       </div>
     </div>
   </div>
+  <ToolDetails
+    ref="toolDetailsRef"
+  />
```
marcelofukumoto: **Consider using the shell's slide-in API instead of a hand-rolled panel.**
useShell().slideIn.open()

We could fix a lot of accessibility problem through this one.

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
```diff
@@ -0,0 +1,376 @@
+<script setup lang="ts">
+import { ref, nextTick } from 'vue';
+import { useStore } from 'vuex';
+import { useI18n } from '@shell/composables/useI18n';
+import { createFocusTrap, FocusTrap } from 'focus-trap';
```
marcelofukumoto: I wonder if we should use the @shell/composables/focusTrap instead since we don't have focus-trap directly on the extension.

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
```diff
@@ -0,0 +1,376 @@
...
+import { UITool } from '../../../../../types';
+import Preview from './Preview.vue';
+
+const store = useStore();
+const { t } = useI18n(store);
+
+const BASE_PATH = 'https://raw.githubusercontent.com/rancher/rancher-ai-ui/main/assets/ui-tools/screenshots';
```
marcelofukumoto: **This fetches from the public internet at runtime.** In air-gapped or egress-restricted installs — a large share of Rancher deployments — every preview fails and the panel shows broken images with no fallback.

It also pins to `main`, so the panel's content isn't versioned with the release it ships in: editing a screenshot on `main` retroactively changes what every released version displays.

Suggest bundling the PNGs as extension assets and resolving them locally. (I see the PR description already flags `BASE_PATH` as a TODO — noting it here as blocking rather than a nit.)

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
```diff
@@ -0,0 +1,376 @@
...
+  isActive.value = true;
+}
+
+function onLeave() {
+  isActive.value = false;
+}
+
+function activateFocusTrap() {
+  nextTick(() => {
+    const slideInElement = document.getElementById('tool-slide-in-content-element');
```
marcelofukumoto: Hardcoded DOM id — breaks if this component is ever mounted more than once (duplicate ids, and `getElementById` picks the first). A template ref on the `<aside>` avoids both and removes the need for the `nextTick` lookup.

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
```diff
@@ -0,0 +1,376 @@
...
+    }
+  });
+}
+
+function deactivateFocusTrap() {
+  if (focusTrap.value) {
+    focusTrap.value.deactivate();
+    focusTrap.value = null;
+  }
+}
```
marcelofukumoto: **The focus trap is never deactivated on unmount.** `deactivateFocusTrap` only runs from `hide()`.

The glass starts at `top: 54px`, so the Rancher top header stays clickable while the panel is open. Navigate from there and this component unmounts with an **active trap bound to a detached node**, leaving document-level listeners that keep grabbing focus.

```js
onBeforeUnmount(deactivateFocusTrap);
```

…plus a route watcher that closes the panel. `SlideInPanelManager` does both (`closeOnRouteChange` + `onBeforeUnmount(closePanel)`).

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
```diff
@@ -0,0 +1,376 @@
...
+      @after-enter="onEnter"
+      @after-leave="onLeave"
+    >
+      <aside
+        v-show="showSlideIn && tool"
+        id="tool-slide-in-content-element"
+        class="slideIn"
+        :class="{'active': isActive}"
+        @keydown.esc="hide()"
+      >
```
marcelofukumoto: **The panel behaves as a modal but isn't announced as one.** It traps focus and has a click-catching backdrop, yet it's an `<aside>` — a *complementary landmark*, semantically the opposite of a modal. There's no `role="dialog"`, no `aria-modal="true"` and no `aria-labelledby` pointing at the `h2`.

Because nothing is `inert`/`aria-modal`, a screen-reader user can still virtual-cursor the entire page behind the panel while sighted keyboard focus is locked inside it.

```html
<aside
  role="dialog"
  aria-modal="true"
  :aria-labelledby="titleId"
  ...
>
```

WCAG 4.1.2 / 1.3.1 (A).

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
```diff
@@ -0,0 +1,376 @@
...
+              <div class="slideIn__header__buttons">
+                <div
+                  class="slideIn__header__button"
+                  role="button"
+                  tabindex="0"
+                  @click="hide()"
+                  @keydown.enter.space="hide()"
+                >
+                  <i class="icon icon-close" />
+                </div>
```
marcelofukumoto: **The close control has no accessible name** — and it's the only control in the panel.

`<div role="button">` containing just `<i class="icon icon-close" />`: an icon font contributes no text, so this announces as bare "button". Use a real `<button>` with a label:

```html
<button
  type="button"
  class="slideIn__header__button"
  :aria-label="t('generic.close')"
  @click="hide()"
>
  <i class="icon icon-close" aria-hidden="true" />
</button>
```

A native `<button>` also gets Enter/Space for free, so the `@keydown.enter.space` handler goes away. (As written that handler does fire on either key — Vue's `withKeys` is OR — but Space will also scroll the page on the way out, so it wants `.prevent`.)

WCAG 4.1.2 (A).

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
```diff
@@ -0,0 +1,376 @@
...
+              </div>
+            </div>
+          </div>
+
+          <div class="tool-category">
+            <i class="category-icon icon icon-category-alt" />
```
marcelofukumoto: Decorative icon-font elements should carry `aria-hidden="true"` — some screen readers announce the private-use glyph. Applies to `icon-category-alt` here, `icon-close` (L109), `icon-version-alt` on the card, and `icon-spinner` in `Preview.vue`.

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
```diff
@@ -0,0 +1,376 @@
...
+            <span class="tool-revision">
+              v{{ tool?.revision }}
+            </span>
+          </div>
+
+          <div class="tool-detail-section">
+            <h3>{{ t('aiConfig.form.section.tools.details.detail') }}</h3>
+          </div>
+
+          <div class="tool-scrollable-content">
```
marcelofukumoto: **This is the root cause of the broken tabbing.**

Count the tabbable nodes inside the trapped container: the close `div[tabindex="0"]` and nothing else. The headings, description, category, revision, the two `v-clean-html` spans and both `` previews are all non-focusable. With exactly **one** tabbable node, Tab and Shift+Tab wrap onto the close button and appear to do nothing.

Knock-on effect here specifically: `.tool-scrollable-content` is an `overflow-y: auto` region whose children contain nothing focusable, so **a keyboard-only user cannot scroll past the fold** — they can open the panel and never reach the usage text or the second screenshot. Chrome's implicit keyboard-focusable-scrollers would normally rescue this, but the `tabbable` library focus-trap uses doesn't recognise it, so the trap actively blocks it.

Fixes: make the previews real buttons (see `Preview.vue`), and give the scroll region `tabindex="0"` with an accessible name (`role="region"` + `aria-label`) so it's reachable and arrow-scrollable.

WCAG 2.1.1 (A).

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
```diff
@@ -0,0 +1,376 @@
...
+            <div class="tool-detail-section">
+              <span
+                v-clean-html="t(`aiConfig.form.section.tools.details.tools.${ tool?.name }.detail`)"
+                class="detail-content"
+              />
```
marcelofukumoto: Two things on this block:

1. **Invalid nesting.** The stylesheet targets `p` and `a` inside `.detail-content` / `.usage-content` (L353-371), but the host element is a `<span>` — `<p>` inside `<span>` is invalid HTML and the parser will break it out, so those rules won't apply as written. Use a `<div>`.

2. **`v-clean-html` isn't buying anything here.** Every one of the new strings is plain prose with no markup, so `{{ }}` would be simpler and strictly safer. If markup in translations is the goal, then point 1 needs fixing first.

3. **No fallback when a tool has no localized copy.** Rancher's `t()` returns `undefined` for a missing key (it doesn't echo the key back), so a tool present in the tools definition but absent from `en-us.yaml` renders a **blank panel with two broken images** — empty title (L94), empty detail, empty usage (L153). The tool catalogue comes from the agent's definition and can change independently of this extension's release, so this is reachable in normal operation, not just in dev. Suggest falling back to the tool's own `description` (already shown on the card) and hiding the preview block when no screenshot exists.

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
```diff
@@ -0,0 +1,376 @@
...
+              <Preview
+                :key="tool?.name"
+                :name="tool?.name"
+                :path="BASE_PATH"
+                :index="0"
+              />
+            </div>
+
+            <div class="tool-usage-section">
+              <h2>{{ t('aiConfig.form.section.tools.details.usage') }}</h2>
```
marcelofukumoto: Heading order skips a level: `h2` (tool name) → `h3` REVISION → `h3` DETAIL → **`h2` Usage**. "Usage" is a sibling of "DETAIL", so it should be `h3`. WCAG 1.3.1 (A).

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
```diff
@@ -0,0 +1,376 @@
...
+</template>
+
+<style lang="scss" scoped>
+  .tool-info-panel {
+    position: fixed;
+    top: 0;
+    left: 0;
+    z-index: 25;
+
+    $slideout-width: 35%;
```
marcelofukumoto: The header offset is hardcoded and slightly off: shell uses `HEADER_HEIGHT = 55` **and** adds `#banner-header`'s height on top. With any banner active (Prime / expiry / custom) the panel and the glass will be misaligned by the full banner height.

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
```diff
@@ -0,0 +1,376 @@
...
+      }
+
+      /* Enter animation */
+      &.slide-enter-active {
+        transition: right 0.5s ease;
+      }
+
+      &.slide-leave-active {
+        transition: right 0.5s ease;
+      }
```
marcelofukumoto: No `prefers-reduced-motion` guard on the 0.5s slide (or on the `fadeIn` keyframes in `Preview.vue`). Cheap to add:

```scss
@media (prefers-reduced-motion: reduce) {
  &.slide-enter-active,
  &.slide-leave-active { transition: none; }
}
```

WCAG 2.3.3 (AAA) — not a release blocker, just easy.

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
```diff
@@ -0,0 +1,376 @@
...
+      font-size: 16px;
+      color: var(--link-text-secondary);
+      margin-right: 8px;
+    }
+
+    .category-value {
+      font-size: 13px;
+      font-weight: 500;
+      color: var(--link-text-secondary);
+      user-select: none;
```
marcelofukumoto: `user-select: none` here and on `.tool-revision` (L324) blocks copying the category and the version — the version being the one string an admin filing a bug would most want to copy out of this panel. Is there a reason to prevent selection?

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/Preview.vue
```diff
@@ -0,0 +1,106 @@
...
+      </div>
+      <img
+        :src="path"
+        :alt="`${props.name} preview`"
+        class="preview-img"
+        :class="{ 'is-loaded': !isLoading }"
+        @load="handleImageLoad"
+        @error="handleImageError"
+        @click="openImageFullScreen"
+      />
```
marcelofukumoto: **The preview is mouse-only.** `@click` on a bare `` with no `tabindex`, no `role` and no key handler — the only way to open a screenshot full-size is a mouse click, and a screen reader won't announce it as interactive at all.

```html
<button type="button" class="preview-button" @click="openImageFullScreen">
  <img :src="path" :alt="..." />
</button>
```

Also worth making the alt text meaningful: both previews for a tool currently get `"<name> preview"`, so they're indistinguishable. Something like `"<tool> — step 1 of 2"`, or `alt=""` + a visible caption if they're decorative.

WCAG 2.1.1 (A) / 4.1.2 (A).

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/Preview.vue
```diff
@@ -0,0 +1,106 @@
...
+
+const isLoading = ref(true);
+
+const handleImageLoad = () => {
+  isLoading.value = false;
+};
+
+const handleImageError = () => {
+  isLoading.value = false;
+};
```
marcelofukumoto: The loading state is silent and the error state is a dead end:

- The spinner has no `role="status"` / `aria-live` and no text alternative, so nothing is announced while previews load.
- `handleImageError` just drops the spinner, leaving a broken image with alt text. Given the images come from `raw.githubusercontent.com` (see `BASE_PATH`), this will fire routinely in air-gapped installs — it wants a real fallback (hide the block, or an explanatory placeholder).

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/Preview.vue
```diff
@@ -0,0 +1,106 @@
...
+
+const openImageFullScreen = () => {
+  window.open(path.value, '_blank');
```
marcelofukumoto: Nit: `window.open(..., '_blank')` without `'noopener,noreferrer'`.

Bigger question — is a new browser tab to a raw GitHub URL the right affordance? It leaves the product for an unstyled image, and won't work at all when the images end up bundled locally. A lightbox / the existing image-viewer pattern may serve better.

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/Preview.vue
```diff
@@ -0,0 +1,106 @@
...
+    max-width: 100%;
+    max-height: 400px;
+    object-fit: cover;
+    width: 100%;
+    height: 100%;
+    cursor: pointer;
+    margin-top: 4px;
+    transform: scaleX(1.06) scaleY(1.09);
+    transform-origin: center;
+  }
```
marcelofukumoto: `object-fit: cover` plus `transform: scaleX(1.06) scaleY(1.09)` crops **and** distorts the screenshot — which undercuts the point of a preview, and makes the UI text inside it harder to read for low-vision users (the reason they'd open it full-size in the first place). `object-fit: contain` with no transform would show the real thing.

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/index.vue
```diff
@@ -408,8 +417,9 @@ const resetToolsToDefaults = () => {
                   v-for="tool in filteredTools"
                   :id="tool.name"
                   :key="tool.name"
-                  :value="tool"
+                  class="tool-card"
```
marcelofukumoto: This class is dead: the same diff **deletes** the `.tool-card` rule from the stylesheet (see the removed block around L649). Either the class or the deletion is unintended.

Note the deleted rule also carried `height: 200px` — without it card height becomes content-driven, so the grid may now be ragged. Worth a look at the rendered grid.

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/index.vue
```diff
@@ -439,6 +451,7 @@ const resetToolsToDefaults = () => {
                       class="toggle-enable-tool"
                       :value="tool.enabled"
                       :disabled="readOnly"
+                      @click.stop
                       @update:value="updateToolEnabled(tool.name, $event)"
                     />
                   </template>
```
marcelofukumoto: **The toggle has no accessible name.** `ToggleSwitch` derives its name from `:aria-label="onLabel"`, and no `onLabel` is passed — so every switch in the grid announces as an unlabelled switch with no indication of *which tool* it enables. Pre-existing, but it's directly in the tab path of the feature under review and the diff touches this element.

Pass the tool name, e.g. `:on-label` / an explicit `aria-label`.

Separately: `@click.stop` and `<rc-item-card-action>` are two different mechanisms for the same job. `RcItemCard._handleCardClick` already ignores clicks inside `[item-card-action]`, which is what `<rc-item-card-action>` adds — that's the documented contract, and it's what the footer button uses. Wrapping the toggle in `<rc-item-card-action>` and dropping both `.stop`s (the footer one is now redundant) would make this consistent.

WCAG 4.1.2 (A).

pkg/rancher-ai-ui/l10n/en-us.yaml
```diff
@@ -522,6 +522,29 @@ aiConfig:
               tooltip:
                 label: Category
                 action: Filter tools by this category
+        details:
+          revision: 'REVISION'
+          detail: 'DETAIL'
+          usage: 'Usage'
+          tools:
```
marcelofukumoto: Hardcoding per-tool prose in the extension's l10n couples UI releases to the agent's tool catalogue: a tool added or renamed on the agent side ships with an empty details panel until this file catches up (see the fallback comment on `tool-details/index.vue`).

Has sourcing `detail` / `usage` from the tool definition itself been considered, with these strings as the fallback rather than the only source?

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/__tests__/UIToolsConfig.test.ts
```diff
@@ -26,6 +26,9 @@ jest.mock('vuex', () => {
 // Mock i18n with proper parameterized translation support
 jest.mock('@shell/composables/useI18n', () => ({ useI18n: () => ({ t: () => '' }) }));
 
+// Mock useShell from @shell/apis
+jest.mock('@shell/apis', () => ({ useShell: jest.fn(() => ({ slideIn: { open: jest.fn() } })) }));
```
marcelofukumoto: Stale mock — the component under test no longer uses `useShell`; the slide-in is hand-rolled now. (It's also the clue that an earlier iteration used the shell slide-in API — see my comment on `index.vue`.)

cypress/e2e/tests/features/settings/settings-ui-tools-config.spec.ts
```diff
@@ -33,37 +33,13 @@ describe('Feature: settings-ui-tools-config', () => {
...
-      uiToolsConfig.intro().infoBanner().should('be.visible');
-      uiToolsConfig.intro().actionButton().should('contain.text', 'Install UI Tools');
-      uiToolsConfig.toolsGrid().should('not.exist');
-
-      cy.get('[data-testid="rancher-ai-ui-settings-tools"]').screenshot('settings-ui-tools-config-test-2-install-banner');
-    });
-  });
-
-  it('Test 3: Enable/disable the UI Tools toggle and save', () => {
+  it('Test 2: Enable/disable the UI Tools toggle and save', () => {
```
marcelofukumoto: This drops the whole `Install UI Tools banner (tools not installed)` describe block, which was the only coverage of the not-installed path, with no replacement and no mention in the PR description. Intentional?

Also, `enabledCheckbox().set(false)` / `set(true)` became `set()` in both places (L42 and L54), so the test now *toggles* rather than asserting a target state — it depends on the entry state and reads less clearly. Was the explicit-value signature removed?

cypress/e2e/tests/features/settings/settings-ui-tools-config.spec.ts
```diff
@@ -170,6 +146,36 @@ describe('Feature: settings-ui-tools-config', () => {
...
+    // Wait for slide-in panel to appear and be visible
+    cy.get('[data-testid="rancher-ai-ui-tool-info-panel-detail"]').should('be.visible');
+
+    // Verify tool details are displayed in the slide-in
+    cy.contains('View the complete YAML').scrollIntoView().should('be.visible');
+
+    // Click outside the slide-in panel to close it
+    cy.get('body').realClick({
+      x: 200,
+      y: 200
```
marcelofukumoto: A couple of things on the new test:

- `cy.contains('Usage')` matches **any** element on the page containing that word — brittle, and it asserts before the panel-visible check right below it. Prefer the `data-testid` you already have.
- Closing via `cy.get('body').realClick({x: 200, y: 200})` depends on the invisible glass overlay covering that coordinate — a layout change silently turns this into a false pass.
- Given this PR introduces a focus-trapping modal, the valuable assertions to add are the accessibility ones: focus moves into the panel on open, Escape closes it, and focus returns to the originating card.

cypress/support/commands/screen.ts
```diff
@@ -0,0 +1,40 @@
...
+  // Remove Cypress UI elements and fix layout
+  cy.window().then((win) => {
+    // Access parent window (Cypress runner)
+    const runnerDoc = win.parent?.document || win.top?.document;
+
+    if (runnerDoc) {
+      // Fix root element to take full width
+      const root = runnerDoc.querySelector('#resizable-panels-root');
+
+      (root as HTMLElement).style.setProperty('width', '100vw', 'important');
```
marcelofukumoto: Unguarded null deref — `querySelector` can return `null` and this throws immediately, unlike `unifiedRunner` below which *is* guarded. It'll blow up whenever the runner chrome isn't present.

```js
if (root) {
  (root as HTMLElement).style.setProperty('width', '100vw', 'important');
}
```

scripts/collect-screenshots.sh
```diff
@@ -0,0 +1,26 @@
...
+# Example: sh scripts/collect-screenshots.sh cypress/e2e/tests/global-ui/screenshots cypress/collected-screenshots
+
+SOURCE_DIR="$1"
+OUTPUT_DIR="$2"
+
+[ -z "$SOURCE_DIR" ] && { echo "Usage: sh $0 [sourceDir] [outputDir]"; exit 1; }
+[ -z "$OUTPUT_DIR" ] && { echo "Usage: sh $0 [sourceDir] [outputDir]"; exit 1; }
+
+# Create or empty output directory
+rm -rf "$OUTPUT_DIR"/* 2>/dev/null
```
marcelofukumoto: This is invoked with `assets/ui-tools/screenshots`, so it **deletes the committed screenshots first** and then moves in only the ones that were generated successfully (`! -name "*fail*"`).

One failing spec therefore silently removes that tool's screenshots from the working tree — recoverable via git, but a nasty surprise. Safer to collect into a temp dir and swap it in only after the whole run succeeds.

marcelofukumoto (commented):
Overall is good in terms of usability and code.

I am doing more checks now in terms of accessibility.
As soon as I click, ESC does work but there is only one focusable in the entire drawer.
So I can't really scroll and I believe I cant click on the images as well.

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/index.vue
torchiaf: Apparently we have those 2 types of drawer at least.
, we tried with the shell's panel first, but later we decided to use the extension's panel style.

- The shell slidein is used for the resource' details page. We didn't want to use that because it covers the setting's header and it applies a grey-out effect on the rest of the page:

image

- The extension's slide-in is completely custom. We are using that approach in the current PR. Ideally we should implement such slide-in panel in the shell as a new component, then use it in both Extensions page and Liz's ui-tools:

image

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
torchiaf: That approach is buggy apparently - We are using the 'focus-trap' imports here to avoid problems after the first opening.

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
torchiaf: I agree on the versioning - this should be by release name - working on it.
For air-gaped envs, yes, we have that problem but I'm more concerned about the bundle's size, for now they takes like 1.3 MB...

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
torchiaf: Done

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
torchiaf: Done

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
torchiaf: It's a copy-paste from the extension's slide-in panel, fixed. This should be a ported in rancher/dashboard as well...

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
torchiaf: Done

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
torchiaf: Done

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/index.vue
torchiaf: Done

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/Preview.vue
torchiaf: Done

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/Preview.vue
torchiaf: Done

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/Preview.vue
torchiaf: It's to remove the image's border and show only the real content.

cypress/e2e/tests/features/settings/settings-ui-tools-config.spec.ts
torchiaf: It was a duplicate - we are already testing it in another spec.

scripts/collect-screenshots.sh
torchiaf: Agree, it's a nice improvement.

torchiaf:
Thanks for the review @marcelofukumoto 
I had completely ignored accessibility, and this is a point that we need to return on - we have many areas where the accessibility is missing.

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/Preview.vue
```diff
@@ -0,0 +1,165 @@
...
+  window.open(path.value, '_blank');
+};
+</script>
+
+<template>
+  <div
+    class="preview-section"
+    @click="openImageFullScreen"
+    @keydown.enter.stop="openImageFullScreen"
+  >
```
marcelofukumoto: Usual change to button if it is going to have the overall button experience and the click and all interaction.

```suggestion
  <button
    type="button"
    class="preview-section"
    :aria-label="t('aiConfig.form.section.tools.details.image-preview.placeholder', { toolName: props.name, pos: props.index + 1 }, true)"
    @click="openImageFullScreen"
  >
```

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/Preview.vue
```diff
@@ -0,0 +1,165 @@
...
+    <div class="img-container">
+      <div
+        v-if="isLoading"
+        class="spinner-wrapper"
+        role="status"
+        aria-live="polite"
+      >
+        <i class="icon icon-spinner icon-spin icon-3x" />
+        <span style="position: absolute; width: 1px; height: 1px; padding: 0; margin: -1px; overflow: hidden; clip: rect(0, 0, 0, 0); white-space: nowrap; border-width: 0;">Loading preview...</span>
```
marcelofukumoto: Remove the hardcoded text.

```suggestion
        <span class="sr-only">{{ t('aiConfig.form.section.tools.details.image-preview.loading') }}</span>
```

marcelofukumoto (commented):
Overall quite good improvements.
Simple details now, I would even believe not even worth it, but nice to give a final recheck.

Considering you are using that focus-trap. Maybe worth to add as a dependency directly to package.json?

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/Preview.vue
torchiaf: Applied

pkg/rancher-ai-ui/pages/settings/sections/ui-tools-config/tool-details/Preview.vue
torchiaf: Applied

torchiaf:
I pushed your suggestions.
For the `focus-trap`, I added the dependency to the package.json but it didn't update the lock file - that's because the dependency is already included in the `shell` and imported from there.
I think we don't need to add it to the package.json.

Issue #190, opened 2026-04-01
Support UI tools


torchiaf:
We want to implement a mechanism to make the Agent select the correct UI elements to show in the Liz chat, based on the user's requests.

For instance:

- user: 'Show me the list of pods in the default namespace'
- agent: {The response} + the list of UI tools = [show-yaml, navigate-to-pod-list, etc.]

deepakpunia-suse:
**Issue:** While verifying this feature, I encountered a bug where the **"Tools"** accordion in the confirmation dialog renders completely blank.

**Description:**

When executing actions that trigger UI tools (such as `show-yaml`) during the confirmation phase, the frontend displays the "Tools" accordion header. However, upon expanding it, the container is empty and the expected tool UI components (buttons) fail to render.

### Environment Details

* **Rancher:** `v2.14.1`
* **Rancher-UI-Extension:** `1.0.1-rc.2`
* **Rancher-AI-Agent:** `v1.0.2-rc.2`
* **Rancher-MCP-Server:** `v1.0.2-rc.1`

**Attachment:**
Please find the attached screen recording for reference.

https://github.com/user-attachments/assets/bbf5feb8-7532-4774-a303-051539b44250

deepakpunia-suse:
### 🧪 **Feature Verification: SUCCESSFUL**

**Environment Details**
* **Rancher:** v2.14.1
* **Rancher-UI-Extension:** 1.0.1-rc.3
* **Rancher-AI-Agent:** v1.0.2-rc.2
* **Rancher-Mcp-Server:** v1.0.2-rc.1

### **Verification Steps**

**Explore (Navigation Route)**
* **Prompt Used:** *"Take me to the list of Pods in the default namespace on the local cluster."*
* **Result:** Liz successfully generated a navigation button. Clicking it correctly routed the browser directly to the Pods list in the Rancher dashboard.


**Show YAML**
* **Prompt Used:** *"Show me the YAML for the rancher deployment in the cattle-system namespace."*
* **Result:** Verified that the "Show YAML" button appeared and successfully opened the Raw YAML viewer for the specified resource.


**Show YAML Diff**
* **Prompt Used:** *"Scale the coredns deployment in kube-system to 2 replicas, but don't apply it yet. Show me the diff."*
* **Result:** Verified the implementation of the diff viewer; clicking the button rendered a clear side-by-side comparison of the proposed configuration changes.


* **Console Logs**
[...]

PR #212, opened 2026-05-13
Fix UI tools validation: show-yaml and show-yaml-diff

torchiaf:
Contributes to https://github.com/rancher/rancher-ai-ui/issues/190

This fixes 2 issues:

- Some resources are not namespaced, so when the user asks for creating a new namespace for instance, the show-yaml tool validation fails because the `resourceNamespace` input is missing.
  - We are removing the namespace from the show-yaml and show-yaml-diff validations.
  
- The tools are validated individually, so it can happen that the agent dispatches more than 1 tools but all of them are invalid. In this scenario, we display the title + empty list -> so only the title is visible.
  - We are now hiding the whole Tool's container (and the title too) when no Tool component is rendered.


pkg/rancher-ai-ui/components/tools/index.vue
```diff
@@ -97,12 +101,17 @@ const tools = computed(() => {
...
+  gap: 2px;
 }
 
-.chat-msg-tool-tags {
-  display: flex;
-  flex-direction: column;
-  gap: 2px;
+// Hide container (and the title too) if there are no tools to show
+.chat-tools-container {
+  &:not(:has(.chat-msg-tools-container > *)) {
```
marcelofukumoto, line 109: **Comment**
Something if off on this code.
Currently you have: 
```
    v-if="tools.length"
```
So technically this code will never run since ideally the block will not be there if the tools.length is empty? Maybe this is redundant?
It would run on an condition where the all tools returns nothing, but if that is the edge case, the way the code works seems confuse, the tools should be filtered instead before sending to the Tools component for example.

pkg/rancher-ai-ui/ui-tools.json
```diff
@@ -44,8 +44,7 @@
           },
           "resourceNamespace": {
             "type": "string",
-            "description": "The namespace where this ONE resource exists.",
-            "required": true
+            "description": "The namespace where this ONE resource exists. Required only if the resource is namespaced."
```
marcelofukumoto, line 47: **Question**
Is `namespaced` a known name? I mean, it is understandable and I would vote that it is ok to use, so mostly asking.

marcelofukumoto (commented):
I executed some tests and was able to see the differences of the results on resources that are not namespaced. That is good. Just some comments before we merge.

pkg/rancher-ai-ui/ui-tools.json
torchiaf: Yes, it worked in our tests - the lllm is able to distinguish what resources have a namespace (Workloads, pods,..) vs the others (Namespace, Cluster...)

pkg/rancher-ai-ui/components/tools/index.vue
torchiaf: The css rule is applied when the tools.length > 0 but all tools failed their internal validation. 
We don't want to filter the tools list beforehand since we would move the validation code in 1 place for all tools  and pass a lot of input props to the single tools - consider that the rule is only to hide the Tools header I think this is acceptable

marcelofukumoto (approved):
LGTM

PR #215, opened 2026-05-19
Fix scroll behavior

torchiaf:
When the message contains UI tools, the scroll-to-bottom behavior fires before the ui tools are fully rendered, so it stops before reaching the bottom of the chat messages view.

The strategy is to scroll to bottom as soon as:

- The HTML inside the last message changes (no matter what is the content and how is rendered).
- The processing phase changes
- The user open an old chat (wait until all messages are loaded before scrolling down)
- There is an error message

### Screenshots

Before

https://github.com/user-attachments/assets/b654226a-504f-481b-8f77-2be059c47b8d

After

https://github.com/user-attachments/assets/35b85bab-abd1-490b-a75e-428cb32e9f08


pkg/rancher-ai-ui/components/panels/Messages.vue
```diff
@@ -57,8 +55,19 @@ const emit = defineEmits(['update:message', 'confirm:message', 'send:message']);
...
 const messagesView = ref<HTMLDivElement | null>(null);
 
+const lastMessageContainer = ref<HTMLDivElement | null>(null);
+const lastMessageObserver = ref<MutationObserver | null>(null);
+
+// Ref callback to assign the last message container for auto-scrolling for each message/error list
+const containerRef = (count: number, index: number) => {
+  return (elem: any) => {
+    if (index === count - 1) {
+      lastMessageContainer.value = elem?.$el || elem as HTMLDivElement;
```
torchiaf: We are registering always the last message container so that we can observe the HTML changes (UI tools, text, resource buttons rendering)

pkg/rancher-ai-ui/components/panels/Messages.vue
```diff
@@ -107,57 +116,52 @@ function getMessageTemplate(component: MessageTemplateComponent) {
...
-    nextTick(() => {
-      // Auto scroll only if enabled or if NEW user messages are added
-      const doScroll = autoScrollEnabled.value || (oldMsgs && newMsgs && newMsgs.length > oldMsgs.length && newMsgs[newMsgs.length - 1].role === Role.User);
+// Scroll when the last message/error changes (HTML content update)
+watch(
+  lastMessageContainer,
+  (newContainer) => {
+    // Clean up old observer
+    if (lastMessageObserver.value) {
+      lastMessageObserver.value.disconnect();
```
marcelofukumoto: **Suggestion**
I see you have this one, but would be nice to add this on unmount as well to make sure it will be gne.

pkg/rancher-ai-ui/composables/useScrollComposable.ts
```diff
@@ -33,17 +33,33 @@ export function useScrollComposable(containerRef: Ref<HTMLDivElement | null>, la
     lastScrollPosition.value = container.scrollTop;
   }
 
-  const handleScroll = throttle(updateScrollState, 100);
```
marcelofukumoto, line 36: **Question**
Are you sure it is required to remove this one?

pkg/rancher-ai-ui/components/panels/Messages.vue
```diff
@@ -57,8 +55,19 @@ const emit = defineEmits(['update:message', 'confirm:message', 'send:message']);
 
 const messagesView = ref<HTMLDivElement | null>(null);
 
+const lastMessageContainer = ref<HTMLDivElement | null>(null);
+const lastMessageObserver = ref<MutationObserver | null>(null);
+
+// Ref callback to assign the last message container for auto-scrolling for each message/error list
+const containerRef = (count: number, index: number) => {
+  return (elem: any) => {
```
marcelofukumoto: **Nit**
It would be better to type this one. Something like: `HTMLDivElement | null` or `ComponentPublicInstance | HTMLDivElement | null ` wouldn't work?
You have as HTMLDivElement already there. So at least the first one should be ok.

cypress/e2e/po/ui-tools/tool.po.ts
```diff
@@ -0,0 +1,27 @@
+import ComponentPo from '@rancher/cypress/e2e/po/components/component.po';
+
+export default class ToolPo extends ComponentPo {
+  suggestions(index: number) {
+    return this.self().get(`[data-testid="rancher-ai-ui-chat-message-list-option-${ index }"]`);
+  }
+
+  selectOption(index: number) {
+    return this.self().get(`[data-testid="rancher-ai-ui-chat-message-list-option-${ index }"]`);
+  }
```
marcelofukumoto, line 10: **Suggestion**
I think they are the same.

pkg/rancher-ai-ui/components/panels/Messages.vue
```diff
@@ -107,57 +116,52 @@ function getMessageTemplate(component: MessageTemplateComponent) {
...
-    });
-  },
-  {
-    immediate: true,
-    deep:      true
-  }
-);
+    // Setup new observer on the last message container
+    if (newContainer) {
+      lastMessageObserver.value = new MutationObserver(() => {
```
marcelofukumoto: **Nit**
Maybe debouncing this would make sense? To avoid triggering many times when receiving multiple messages.

marcelofukumoto:
Executed the CODE, it fixed the scrolling on new messages! That was cool!
LGTM good job. I will wait for the small code changes.

pkg/rancher-ai-ui/components/panels/Messages.vue
torchiaf: Yes, absolutely

pkg/rancher-ai-ui/composables/useScrollComposable.ts
torchiaf: Yes, the ScrollButton visibilty depends on this function, so when you scroll down fast or moving around {bottom - 150 pixels} - which is the threshold for showing the Scroll button - it causes some flickering because it fires after the 100 ms.


https://github.com/user-attachments/assets/84ade473-83d0-4417-a34e-1579b9022ce6

cypress/e2e/po/ui-tools/tool.po.ts
torchiaf: They are but it's fine - both tools use ListOptions.vue

pkg/rancher-ai-ui/components/panels/Messages.vue
torchiaf: Mm this should be already prevented here right? Even when loading a previous chat, it should fires only on last message.

```
  if (index === count - 1) {
    lastMessageContainer.value = (elem as ComponentPublicInstance)?.$el || elem;
  }
```

lastMessageContainer.value changes only on last message.

marcelofukumoto: What if receives many message and keep changing it?

torchiaf: Alright let's try with a 100 ms debounce time - It seems that it doesn't break anything but it's a nice to have.

torchiaf: All good - no regressions with the debounce

pkg/rancher-ai-ui/components/message/index.vue
```diff
@@ -123,7 +123,6 @@ function handleToolAction(event: ToolActionEvent) {
         >
           <BubbleButton
             v-if="props.message.role === RoleEnum.Assistant && !!props.message.thinkingContent"
-            data-testid="rancher-ai-ui-chat-message-show-thinking-button"
```
marcelofukumoto, line 126: **Comment**
Removed testid. Any reason? Not using?

marcelofukumoto (approved):
Tested again. LGTM

PR #216, opened 2026-05-20
System suggestions refactoring

torchiaf:
It fixes some inconsistencies in system suggestions messages -> we want to make them generic in sight of OAuth2 auth feature.

### How to test

We have to verify that the message below (system message) works fine - it is displayed after 5+ user requests sent using Adaptive mode.

image

### DIFF
--- a/pkg/rancher-ai-ui/components/message/template/SystemRequest.vue
+++ b/pkg/rancher-ai-ui/components/message/template/SystemRequest.vue
@@ -41,35 +41,35 @@ function doAction(type: 'confirm' | 'cancel') {
 
 <template>
   <div
-    class="chat-system-suggestion-message"
+    class="chat-system-request-message"
     :class="{ 'disabled-panel': props.disabled }"
   >
     <SystemAvatar class="chat-msg-avatar" />
     <div
       v-if="props.message?.templateContent?.content?.message"
-      class="chat-system-suggestion-msg-bubble"
+      class="chat-system-request-msg-bubble"
     >
-      <div class="chat-system-suggestion-msg-text">
+      <div class="chat-system-request-msg-text">
         <span
           v-clean-html="formatMessageContent(props.message.templateContent.content.message || '')"
         />
       </div>
 
-      <div class="chat-system-suggestion-actions">
+      <div class="chat-system-request-actions">
         <div
           v-if="props.message.confirmation"
           :data-testid="`rancher-ai-ui-chat-message-confirmation-${props.message.confirmation.status}`"
-          :class="['chat-system-suggestion-actions-result', `status-${props.message.confirmation.status}` ]"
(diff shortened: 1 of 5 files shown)


pkg/rancher-ai-ui/types.ts
```diff
@@ -185,17 +185,12 @@ export interface MessageConfirmationAction {
...
 
 export interface MessageTemplate {
   component: MessageTemplateComponent;
-  content: MessageTemplateContent;
-}
-
-export interface MessageTemplateContent {
-  principal?: any;
-  message: string;
+  content: any;
```
torchiaf: This is generic now

marcelofukumoto: **Suggestion**
Since we already have an idea of the MessageTemplateContent, wouldn't be better to use something like this instead? At least we have the old one and anything can be added. It would even be better to add the ones we already know that can be used.
```
export interface MessageTemplate {
  component: MessageTemplateComponent;
  content: { message?: string; principal?: any; [key: string]: unknown };
}
```

pkg/rancher-ai-ui/components/message/template/SystemRequest.vue
```diff
@@ -155,17 +155,17 @@ function doAction(type: 'confirm' | 'cancel') {
   }
 }
 
-.theme-dark .chat-system-suggestion-msg-text :deep(code) {
+.theme-dark .chat-system-request-msg-text :deep(code) {
   color: #C0EFDE;
```
marcelofukumoto, line 159: **Nit**
Maybe we can fix this hard-coded color.

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
```diff
@@ -217,23 +218,43 @@ export function useChatMessageComposable(
...
-        content:   { message: t('ai.message.system.switchAgent.info', { agent: agent?.displayName || currentMsg.value.agentMetadata?.recommended }, true) }
+        component,
+        content
       },
-      actions: [
+      actions
+    };
+  }
+
+  async function ensureSwitchAgentSuggestion(agentName = '') {
```
marcelofukumoto, line 239: **Suggestion**
Since we have some changes on functions, it would be nice to add some unit tests here as well.

torchiaf: I agree, I added some tests for `useChatMessageComposable`

pkg/rancher-ai-ui/components/message/template/SystemRequest.vue
torchiaf: Let's tackle this in another PR - we also have to fix the `size` variant and re-consider some CSS rules

marcelofukumoto (approved):
LGTM
Feature still working fine as intended.

PR #218, opened 2026-05-20
Update cron to run every weekday at 1AM UTC. Update version.

marcelofukumoto:
Now that we have defined that the e2e test plan automation is working fine. Update to run daily.


torchiaf (approved):
LGTM

PR #222, opened 2026-05-21
Fix chat panel height

### LINKED ISSUE #221: The Chat panel is cut off when Rancher has a Header banner
When a Header or Footer banner is set up in Rancher, the Chat panel is shifted to the bottom because the panel's height is static

### How to replicate

- Go to Global Settings -> Banners
- Enable the "Header Banner"
- Edit the "Text" field, add a 3 line text
- Save
- Open the Chat

Image

Image

torchiaf:
Fixes https://github.com/rancher/rancher-ai-ui/issues/221

This is fixing the Chat panel and Chat's history panel height to be adaptive to the parent's height

### Screenshots

Before

image

After

image

### DIFF
--- a/cypress/e2e/tests/features/chat.spec.ts
+++ b/cypress/e2e/tests/features/chat.spec.ts
@@ -6,13 +6,13 @@ import UsersPo from '@rancher/cypress/e2e/po/pages/users-and-auth/users.po';
 import ExtensionsPagePo from '@rancher/cypress/e2e/po/pages/extensions.po';
 import ProductNavPo from '@rancher/cypress/e2e/po/side-bars/product-side-nav.po';
 import { SettingsPagePo as GlobalSettings } from '@rancher/cypress/e2e/po/pages/global-settings/settings.po';
-import { BannersPagePo } from '@rancher/cypress/e2e/po/pages/global-settings/banners.po';
 import { SettingsPagePo } from '@/cypress/e2e/po/settings.po';
 import ChatPo from '@/cypress/e2e/po/chat.po';
 import { HistoryPo } from '@/cypress/e2e/po/history.po';
 import RancherHeaderPo from '@/cypress/e2e/po/components/rancher-header.po';
 import { SlidingBadgePo } from '@/cypress/e2e/po/hook.po';
 import ContextPo from '@/cypress/e2e/po/context.po';
+import RancherSettingsBannersPo from '@/cypress/e2e/po/components/rancher-settings-banners.po';
 import ApplySettingsPromptPo from '@/cypress/e2e/po/dialog/apply-settings.po';
 import { rancherAgentConfig, fleetAgentConfig, provisioningAgentConfig } from '@/cypress/e2e/blueprints/aiAgentConfigs';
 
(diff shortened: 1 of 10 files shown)


pkg/rancher-ai-ui/pages/Chat.vue
```diff
@@ -471,7 +471,7 @@ function unmount() {
 .chat-container {
   display: flex;
   flex-direction: row;
-  height: calc(100vh - 55px);
+  height: stretch;
```
marcelofukumoto: **Question**
Just worried about compatibility of STRETCH on older browsers.
I checked it has coverage for all browsers, mostly thinking on older ones.

cypress/e2e/tests/features/chat.spec.ts
```diff
@@ -914,4 +916,153 @@ describe('Chat', () => {
...
+        bannersPage.headerBannerCheckbox().set();
+
+        bannersPage.headerInput().set('test\ntest\ntest');
+
+        bannersPage.footerBannerCheckbox().set();
+
+        bannersPage.footerInput().set('test\ntest\ntest');
+
+        bannersPage.applyAndWait('**/ui-banners', 200);
+
```
marcelofukumoto: **Nit**
This code and top one are same right? Maybe we can turn this into a function instead?

marcelofukumoto (commented):
Overall seems a controlled change. But those are really to bug it out for different sizes.
Specially the change to flex:1 and added display:flex. After the code reviewed I will smoke test the panel.

cypress/e2e/tests/features/chat.spec.ts
torchiaf: Done

pkg/rancher-ai-ui/pages/Chat.vue
torchiaf: I added `height: 100` as fallback for old browsers.

marcelofukumoto:
I tried to replicate a case where this would bug naturally.
I couldn't find a bug.
The only thing I could do is just like you did and add some DIVs on top of it.

**Question**
What bug is this one actually fixing?

torchiaf:
> I tried to replicate a case where this would bug naturally. I couldn't find a bug. The only thing I could do is just like you did and add some DIVs on top of it.
> 
> **Question** What bug is this one actually fixing?

@marcelofukumoto take a look at https://github.com/rancher/rancher-ai-ui/issues/221 - when Rancher has banners (either footer or header's banners) the Chat panel is not adapting to the parent's height -> the Console panel is hidden.

image

The e2e tests are now verifying this scenario: https://github.com/torchiaf/rancher-ai-ui/blob/c2e0d9aa7a054d28b163c9e2500a79df5a67c0dc/cypress/e2e/tests/features/chat.spec.ts#L1079

marcelofukumoto:
Oh that is true!
For some reason I thought you just added at the top something! And I didnt see the actual ticket ! Shame on me.

image

Seems good for chrome, but doesnt work well on Firefox and Safari and Edge

Not sure if there is anything biased on my env. But I would not think so. Maybe related for it been 100%?
Firefox:
image

Safari:
image

Edge:
image

Not sure if is the way I built, when I run as localhost I dont have the problem.

Let me check if I find anything. Or something I am doing wrong.

marcelofukumoto:
Took my a while to rebuild everything to remove.
Tried again on localhost, I face the bug.
Tried with the serve-pkg, same.

So there is a bug there for anything that is not Chrome. Let me know if you need a call for us to check on it.

torchiaf:
> Oh that is true! For some reason I thought you just added at the top something! And I didnt see the actual ticket ! Shame on me.
> 
> Not sure if is the way I built, when I run as localhost I dont have the problem.
> 
> Let me check if I find anything. Or something I am doing wrong.

@marcelofukumoto  you are right - the layout breaks in Edge and Firefox. I fixed it by removing the `height: stretch` totally, `height: 100%` works on all browsers.
Please take a look, thanks!

marcelofukumoto:
I rechecked again. It worked well on SAFARI, FIREFOX and EDGE.
Worked fine on CHROME as well.

marcelofukumoto (approved):
LGTM

PR #223, opened 2026-05-21
Various fixes, ui tools e2e tests

torchiaf:
The ui tools [tests](4aff67727d87d2e14ea6e31a1c0f1c095344423b) found some issues:

- 1a1a5991886f25f5a6f59a9ae5cfc65359eab0c2 : the random key is breaking the references to the props.message in the ShowYaml and ShowYamlDiff. When the list of tools are pre-defined, is safer to use the index as key, otherwise we can continue using the random key

  before - it doesn't close the editor, the ref to the current message status is lost

  https://github.com/user-attachments/assets/af745980-37f1-4875-b67d-fde90f698223

  after - the editor can close

  https://github.com/user-attachments/assets/ffe98583-c86c-4ded-b854-266ea57fd6db

- 2fddd04d80af6df0c8a9e7b246fa869e90b5cee6 : we want to scroll to bottom when opening a previous chat

- 9193e2ce92946b99135010a0fd9d6c83edf346c6 : debounce removed - the 100 ms was breaking the auto scroll when the processing label changes before the last message (Confirmation messages)

  before

    https://github.com/user-attachments/assets/9cdd9ee3-9575-4f69-af49-b922177189b1

  after

   https://github.com/user-attachments/assets/a5a1c130-4008-429b-913b-2410326cb33f

### DIFF
--- a/cypress/e2e/tests/features/chat.spec.ts
+++ b/cypress/e2e/tests/features/chat.spec.ts
@@ -10,6 +10,7 @@ import { SettingsPagePo } from '@/cypress/e2e/po/settings.po';
 import ChatPo from '@/cypress/e2e/po/chat.po';
 import { HistoryPo } from '@/cypress/e2e/po/history.po';
 import { SlidingBadgePo } from '@/cypress/e2e/po/hook.po';
+import ContextPo from '@/cypress/e2e/po/context.po';
 import ApplySettingsPromptPo from '@/cypress/e2e/po/dialog/apply-settings.po';
 import { rancherAgentConfig, fleetAgentConfig, provisioningAgentConfig } from '@/cypress/e2e/blueprints/aiAgentConfigs';
 
@@ -457,6 +458,54 @@ describe('Chat', () => {
       chat.getMessage(3).timestamp().should('be.visible');
 
       chat.scrollButton().checkNotExists();
+    });
+
+    it('it should automatically scroll to bottom when receiving a confirmation message', () => {
+      HomePagePo.goTo();
+
+      chat.open();
+      chat.isReady();
+
+      const welcomeMessage = chat.getMessage(1);
+
+      welcomeMessage.isCompleted();
+
+      cy.enqueueLLMResponse({
+        text:      'Pod creation confirmed.',
+        mcpTool: {
+          name: 'createKubernetesResource',
+          args: {
+            kind:      'Pod',
+            name:      'my-pod',
+            resource:  {
+              apiVersion: 'v1',
+              kind:       'Pod',
+              metadata:   {
+                name:      'my-pod',
+                namespace: 'default'
+              },
+            },
+            cluster:   'local',
+            namespace: 'default'
+          }
+        },
+      });
+
+      chat.sendMessage('Create a pod named my-pod in default namespace');
+
+      const userMessage = chat.getMessage(2);
+
+      userMessage.containsText('Create a pod named my-pod in default namespace');
+
+      // Verify the processing label is visible and not covered by the console panel
+      chat.phase('Awaiting confirmation').then(($phase) => {
+        new ContextPo().self().then(($context) => {
+          const phaseRect = $phase[0].getBoundingClientRect();
+          const contextRect = $context[0].getBoundingClientRect();
+
+          expect(contextRect.top + 2).to.be.greaterThan(phaseRect.top);
+        });
+      });
     });
 
     it('it should not scroll to bottom when message stream is in progress and user scrolls up', () => {
--- a/cypress/e2e/tests/features/chat.spec.ts
+++ b/cypress/e2e/tests/features/chat.spec.ts
@@ -536,6 +536,40 @@ describe('Chat', () => {
       chat.scrollButton().checkNotExists();
     });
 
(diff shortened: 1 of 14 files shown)


cypress/e2e/tests/features/chat.spec.ts
```diff
@@ -535,6 +585,40 @@ describe('Chat', () => {
       chat.scrollButton().checkNotExists();
     });
 
+    it('It should automatically scroll to bottom when close and reopen the chat panel', () => {
```
marcelofukumoto: **Nit**
This should start with small case to follow the pattern from the others.

cypress/e2e/tests/features/ui-tools.spec.ts
```diff
@@ -171,56 +180,769 @@ describe('UI Tools', () => {
...
+
+        yamlEditor.title().should('contain.text', 'Pod');
+        yamlEditor.title().should('contain.text', 'cattle-ai-agent-system/test-pod');
+
+        confirmationRequestMessage.confirmButton().click({ force: true });
+
+        yamlEditor.checkNotExists();
+
+        confirmationRequestMessage.isConfirmed();
+        confirmationRequestMessage.containsText('Confirmed');
```
marcelofukumoto, line 953: **Question**
Is this one checking for CONFIRMED or CANCELED? The title of the test says
`it should close the editor when cancel a confirmation request in the Chat`

But here is checking for Confirmed?

cypress/e2e/tests/features/ui-tools.spec.ts
```diff
@@ -171,56 +180,769 @@ describe('UI Tools', () => {
...
+
+        yamlEditor.title().should('contain.text', 'Pod');
+        yamlEditor.title().should('contain.text', 'cattle-ai-agent-system/test-pod');
+
+        confirmationRequestMessage.cancelButton().click({ force: true });
+
+        yamlEditor.checkNotExists();
+
+        confirmationRequestMessage.isCanceled();
+        confirmationRequestMessage.containsText('Canceled');
```
marcelofukumoto, line 1017: **Comment**
And this one says Canceled. Maybe I am missing something, but from the other test the name is correct.

cypress/e2e/tests/features/ui-tools.spec.ts
```diff
@@ -171,56 +180,769 @@ describe('UI Tools', () => {
...
+
+        resultMessage.isCompleted();
+
+        const showLogs = resultMessage.tool().openConsoleLogs('local', testPod.metadata.namespace, testPod.metadata.name, testPod.spec.containers[0].name);
+
+        showLogs.click();
+
+        cy.get(`[data-testid="horizontal-window-manager"]`).should('contain.text', testPod.metadata.name);
+
+        cy.deleteRancherResource('v1', 'pods', `${ testPod.metadata.namespace }/${ testPod.metadata.name }`, false);
```
marcelofukumoto, line 409: **Suggestion**
Here you have the cy.createRancherResource and the cy.deleteRancherResource inside the same block.
If the TEST fails the DELETE will not run.

I think it is better to move to an BEFORE/AFTER, or at least move the delete to AFTER to make sure it will be deleted.

cypress/e2e/tests/features/ui-tools.spec.ts
```diff
@@ -171,56 +180,769 @@ describe('UI Tools', () => {
...
+
+        const yamlEditor = new YamlEditorPo();
+
+        yamlEditor.title().should('contain.text', 'Pod YAML');
+        yamlEditor.title().should('contain.text', 'default/test-pod');
+
+        yamlEditor.confirmButton().should('not.exist');
+        yamlEditor.cancelButton().should('not.exist');
+
+        yamlEditor.close();
```
marcelofukumoto, line 462: **Suggestion**
Here we have the TITLE/CONFIRM BUTTON/CANCEL BUTTON checks
Maybe would be better to check the CONTENT inside as well?

marcelofukumoto (commented):
Gave some checks :) Seems good though! 
Thanks for adding E2E tests!

cypress/e2e/tests/features/chat.spec.ts
torchiaf: Fixed

cypress/e2e/tests/features/ui-tools.spec.ts
torchiaf: This is a typo, the title is wrong, it should be 'confirm'

cypress/e2e/tests/features/ui-tools.spec.ts
torchiaf: Same of above, the title is wrong, it should be 'cancel'.

cypress/e2e/tests/features/ui-tools.spec.ts
torchiaf: Agree

cypress/e2e/tests/features/ui-tools.spec.ts
torchiaf: Done - also checking the edit buttons when opening the editor in edit mode (yaml-diff)

marcelofukumoto (approved):
LGTM.

PR #232, opened 2026-05-28
Support sub-agents

### LINKED ISSUE #207: Support sub-agents architecture
Related issue: https://github.com/rancher/rancher-ai-agent/issues/169

### LINKED ISSUE #239: Improve welcome message
Related issue: https://github.com/rancher/rancher-ai-agent/issues/209

torchiaf:
Fixes https://github.com/rancher/rancher-ai-ui/issues/207
Fixes https://github.com/rancher/rancher-ai-ui/issues/239

### 1. Support sub-agents architecturre

The AI agent v1.1.0 implements a sub-agents architecture: https://github.com/rancher/rancher-ai-agent/blob/main/ARCHITECTURE.md

Single Agent selection VS Sub agents selection UI changes:

- Since the Agent can now select multiple agents for resolving user's requests, it does make sense to change the message's header to "Adaptive Agent(s) Selection"

  image

- The UI now receives a short description of what Agents are in action; we are displaying those labels as processing labels at the bottom of Message view

  image

- The AgentMetadata is not relevant in the new subagents context, since we don't want to track every single agents involved in the user's request - it's now used only to track the recommended agent info

### 2. Improve Welcome message

In the new architecture, we noticed that the Welcome message generates performance issues.
It sends a request to get the `suggestions` tool but this is involving the Rancher MCP which is not really needed in this context.
We are removing the Welcome message request from the conversation's stream and replacing it with a simple http request that returns the list of suggestions.

### 3. Adapt e2e tests

- The agent now returns faster responses, we are adapting the e2e tests to the new conditions.
- The LLM mock server now supports responses from multiple agents, we changed the control API interface to support the new payload

### DIFF
--- a/cypress/e2e/po/chat.po.ts
+++ b/cypress/e2e/po/chat.po.ts
@@ -1,8 +1,8 @@
 import ComponentPo from '@rancher/cypress/e2e/po/components/component.po';
 import RancherHeaderPo from '@/cypress/e2e/po/components/rancher-header.po';
-import { MessagePo, ErrorMessagePo } from '@/cypress/e2e/po/message.po';
 import { HeaderPo } from '@/cypress/e2e/po/header.po';
 import { ConsolePo } from '@/cypress/e2e/po/console.po';
+import MessagesPo from '@/cypress/e2e/po/messages.po';
 
 export default class ChatPo extends ComponentPo {
   rancherHeader: RancherHeaderPo;
@@ -13,31 +13,19 @@ export default class ChatPo extends ComponentPo {
   }
 
   processingState(label?: string) {
-    if (label) {
-      return this.self().get(`[data-testid="rancher-ai-ui-processing-state-${ label.toLowerCase().replace(/\s/g, '-') }"]`);
-    }
-
-    return this.self().get('[data-testid^="rancher-ai-ui-processing-state-"]');
+    return this.messagesPanel().processingState(label);
   }
 
   header() {
     return new HeaderPo();
   }
 
+  messagesPanel() {
+    return new MessagesPo();
+  }
+
   console() {
     return new ConsolePo();
-  }
-
-  scrollButton() {
-    return new ComponentPo('[data-testid="rancher-ai-ui-scroll-button"]');
-  }
-
-  scrollTop(args?: { wait?: number }) {
-    if (args?.wait) {
-      cy.wait(args.wait);
-    }
-
-    this.getMessage(1).scrollIntoView();
   }
 
   isReady(timeout = 10000) {
@@ -71,15 +59,15 @@ export default class ChatPo extends ComponentPo {
   }
 
   getMessage(id: string | number) {
-    return new MessagePo(id.toString());
+    return this.messagesPanel().getMessage(id);
   }
 
   getErrorMessage(index: number) {
-    return new MessagePo(index.toString());
+    return this.messagesPanel().getErrorMessage(index);
   }
 
   getSystemErrorMessage(index: number) {
-    return new ErrorMessagePo(index.toString());
+    return this.messagesPanel().getSystemErrorMessage(index);
   }
 
   sendMessage(value: string) {
--- a/cypress/e2e/po/chat.po.ts
+++ b/cypress/e2e/po/chat.po.ts
@@ -12,8 +12,12 @@ export default class ChatPo extends ComponentPo {
     this.rancherHeader = new RancherHeaderPo();
   }
 
-  phase(label: string) {
-    return this.self().get(`[data-testid="rancher-ai-ui-processing-phase-${ label.toLowerCase().replace(/\s/g, '-') }"]`);
+  processingState(label?: string) {
+    if (label) {
+      return this.self().get(`[data-testid="rancher-ai-ui-processing-state-${ label.toLowerCase().replace(/\s/g, '-') }"]`);
+    }
+
+    return this.self().get('[data-testid^="rancher-ai-ui-processing-state-"]');
   }
 
   header() {
@@ -26,6 +30,14 @@ export default class ChatPo extends ComponentPo {
 
   scrollButton() {
     return new ComponentPo('[data-testid="rancher-ai-ui-scroll-button"]');
+  }
+
+  scrollTop(args?: { wait?: number }) {
+    if (args?.wait) {
+      cy.wait(args.wait);
+    }
+
+    this.getMessage(1).scrollIntoView();
   }
 
   isReady(timeout = 10000) {
--- a/cypress/e2e/tests/features/multi-agent/chat.spec.ts
+++ b/cypress/e2e/tests/features/multi-agent/chat.spec.ts
@@ -21,6 +21,9 @@ describe('Multi Agent Chat', () => {
     beforeEach(() => {
       cy.login();
 
+      cy.clearLLMResponses();
+      cy.cleanChatHistory();
+
       HomePagePo.goTo();
 
       chat.open();
@@ -35,7 +38,7 @@ describe('Multi Agent Chat', () => {
 
       const selectAgent = chat.console().selectAgent();
 
-      selectAgent.self().contains('Adaptive Agent Selection');
+      selectAgent.self().contains('Adaptive Agent(s) Selection');
 
       selectAgent.open();
 
@@ -54,17 +57,23 @@ describe('Multi Agent Chat', () => {
     });
 
     it('It should be possible to switch agent selection mode from a system recommendation message', () => {
-      // Sending 5 messages to trigger a system recommendation to switch agent selection mode
-      for (let i = 0; i < 5; i++) {
+      // Sending 3 messages to trigger a system recommendation to switch agent selection mode
+      for (let i = 0; i < 3; i++) {
+        cy.enqueueLLMResponse({
+          agentResponses: [{ agent: 'rancher' }],
+          text:           `Response ${ i + 1 }`
+        });
         chat.sendMessage(`Request ${ i + 1 }`);
 
         const responseMessage = chat.getMessage(3 + (i * 2));
 
         responseMessage.isCompleted();
+
+        responseMessage.agentSelectionLabel().contains('Adaptive Agent(s) Selection');
       }
 
       // Confirm the mode switch from the system message
-      const switchAgentMessage = chat.getMessage(12);
+      const switchAgentMessage = chat.getMessage(7);
 
       switchAgentMessage.confirmButton().click();
 
@@ -73,12 +82,12 @@ describe('Multi Agent Chat', () => {
       // Verify that the selection mode has been switched in the console
       const selectAgent = chat.console().selectAgent();
 
-      selectAgent.self().should('not.contain', 'Adaptive Agent Selection');
+      selectAgent.self().should('not.contain', 'Adaptive Agent(s) Selection');
 
       // Verify that the selection mode is reflected in the next messages
       chat.sendMessage('Check current agent');
 
-      const responseMessage = chat.getMessage(14);
+      const responseMessage = chat.getMessage(8);
 
       responseMessage.scrollIntoView();
       responseMessage.containsText('Rancher');
@@ -86,17 +95,24 @@ describe('Multi Agent Chat', () => {
     });
 
     it('It should be possible to dismiss agent switch from a system recommendation message', () => {
-      // Sending 5 messages to trigger a system recommendation to switch agent selection mode
-      for (let i = 0; i < 5; i++) {
+      // Sending 3 messages to trigger a system recommendation to switch agent selection mode
+      for (let i = 0; i < 3; i++) {
+        cy.enqueueLLMResponse({
+          agentResponses: [{ agent: 'rancher' }],
+          text:           `Response ${ i + 1 }`
+        });
+
         chat.sendMessage(`Request ${ i + 1 }`);
 
         const responseMessage = chat.getMessage(3 + (i * 2));
 
         responseMessage.isCompleted();
+
+        responseMessage.agentSelectionLabel().contains('Adaptive Agent(s) Selection');
       }
 
       // Dismiss the mode switch from the system message
-      const switchAgentMessage = chat.getMessage(12);
+      const switchAgentMessage = chat.getMessage(7);
 
       switchAgentMessage.cancelButton().click();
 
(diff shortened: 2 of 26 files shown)


cypress/e2e/po/chat.po.ts
```diff
@@ -31,6 +35,14 @@ export default class ChatPo extends ComponentPo {
     return new ComponentPo('[data-testid="rancher-ai-ui-scroll-button"]');
   }
 
+  scrollTop(args?: { wait?: number }) {
+    if (args?.wait) {
```
torchiaf: The new agent is faster now - we have to introduce a small delay when scrolling up

cypress/e2e/tests/features/multi-agent/chat.spec.ts
```diff
@@ -249,8 +269,8 @@ describe('Multi Agent Chat', () => {
 
       chat.open();
 
-      chat.getSystemErrorMessage(1).containsText('Failed to load MCP tools for all enabled agents.');
-      chat.getSystemErrorMessage(1).containsText('Please check the AI Agents configuration and ensure the MCP server is accessible with the provided connection details.');
+      chat.getSystemErrorMessage(1).containsText(`Failed to load MCP tools for agent 'invalid-agent'`);
```
torchiaf, line 279: The error message has been improved in the backend - this reflects the new message

cypress/support/commands/llm-mock-service-api.ts
```diff
@@ -132,10 +129,8 @@ Cypress.Commands.add('enqueueLLMResponse', (args: LlmResponseArgs) => {
     }
   }
 
-  // We are in a multi-agent environment, select the default agent if not specified.
-  const agent = args.agent === undefined ? 'rancher' : args.agent;
+  const agentResponses = args.agentResponses || [];
```
torchiaf, line 132: The llm mock server now accepts a MCP tool call for each Agent mocked in the request, see https://github.com/rancher-sandbox/rancher-ai-llm-mock/blob/main/openapi.yaml

pkg/rancher-ai-ui/components/Processing.vue
```diff
@@ -41,11 +48,15 @@ const dots = computed(() => {
 const phase = ref(props.phase);
 const applyPhase = debounce((v) => {
   phase.value = v;
-}, 150);
 
-watch(() => props.phase, (v) => applyPhase(v), { immediate: true });
+  nextTick(() => emit('change:phase', v));
```
torchiaf, line 52: The agent is now really fast - we need to send the new phase event exactly when it changes. Using the event here also allow to remove the phase's watcher in the Messages component, which is good.

pkg/rancher-ai-ui/utils/format.ts
```diff
@@ -111,24 +113,16 @@ export function formatAgentFromCRD(config: AIAgentConfigCRD): Agent {
...
-      const agent = agents.find((a) => a.name === agentName);
-
-      if (agent) {
-        return {
-          agent,
-          selectionMode,
-          recommended,
-        };
-      }
+      return { recommended };
```
torchiaf, line 125: We only need the recommended Agent now, it's required for displaying the Switch Agent message.

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
```diff
@@ -399,63 +366,78 @@ export function useChatMessageComposable(
     }
   }
 
-  async function processWelcomeData(data: string) {
-    switch (data) {
-    case Tag.MessageStart:
-      const msgId = await addMessage(buildWelcomeMessage());
+  async function ensureWelcomeMessage() {
```
torchiaf, line 369: This now builds the Welcome message by fetching the `suggestions` via http call

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
```diff
@@ -464,13 +446,29 @@ export function useChatMessageComposable(
...
-      if (data.startsWith(Tag.AgentMetadataStart) && data.endsWith(Tag.AgentMetadataEnd)) {
-        const metadata = formatAgentMetadata(data, agents.value);
+      if (data.startsWith(Tag.ProcessingSubagentInitStart) && data.endsWith(Tag.ProcessingSubagentInitEnd)) {
+        const metadata = formatSubAgentProcessingMetadata(data, agents.value);
 
         if (metadata) {
-          currentMsg.value.agentMetadata = metadata;
+          setProcessingState({
+            phase: MessagePhase.ProcessingSubagent,
+            label: metadata.query ? `${ metadata.agent?.displayName || metadata.agent || '' }: ${ metadata.query.slice(0, 100) }` : ''
```
marcelofukumoto: **Suggestion**
1. Is this is the  metadata.agent. I was just wondering since metadata.agent is actually and object. \\And if 2. There is a guard case to just have '', but than the result would be ": metadata.query.slice", so if this is the case, better to guard the ":" as well.

pkg/rancher-ai-ui/composables/__tests__/useChatMessageComposable.test.ts
```diff
@@ -1,559 +1,1217 @@
...
+  });
 
-    it('should handle message end tag in welcome', async() => {
-      // When data = Tag.MessageEnd = '</message>'
-      // processWelcomeData should:
-      // 1. Set phase to MessagePhase.Idle
-      // 2. Clear messageContent
-      // 3. Set completed to true
+  describe('Chat initialization', () => {
+  });
```
marcelofukumoto: **Suggestions**
Is this empty?

cypress/e2e/po/chat.po.ts
marcelofukumoto: **Nit**
Since it is a scroll, we can check for some components showing up. Nothing major since just a scroll.

pkg/rancher-ai-ui/l10n/en-us.yaml
```diff
@@ -256,17 +257,16 @@ ai:
         delete: Delete Chat
   agents:
     selectAgent:
-      tooltip: 'Select an Agent to handle your request or use "Adaptive Agent Selection" to let the system choose the best agent based on your context and needs.'
+      tooltip: 'Select an Agent to handle your request or use "Adaptive Agent(s) Selection" to let the system choose the best agent based on your context and needs.'
```
marcelofukumoto, line 263: **Question**
Why we added the (s)? Because we have many Agents we can select? If it is because of the added sub-agents, than it doesnt feel right because agent and sub-agents are different.
Just curious since it is a really specific change. :)

pkg/rancher-ai-ui/components/Processing.vue
```diff
@@ -57,6 +68,8 @@ onMounted(() => {
   interval = setInterval(() => {
     now.value = Date.now();
   }, 500);
+
+  watch(() => props.phase, (v) => applyPhase(v), { immediate: true });
 });
 
 onBeforeUnmount(() => {
```
marcelofukumoto, line 75: **Suggestion**
Maybe canceling the applyPhase when unmount just to not trigger on unmount.

pkg/rancher-ai-ui/store/chat.ts
```diff
@@ -190,14 +190,14 @@ const mutations = {
...
-    const { chatId, phase } = args;
+  setProcessingState(state: State, args: { chatId: string; processingState: MessageProcessingState }) {
+    const { chatId, processingState } = args;
 
     if (!chatId || !state.chats[chatId]) {
       return;
     }
 
-    state.chats[chatId].phase = phase;
+    state.chats[chatId].processingState = processingState;
```
marcelofukumoto, line 200: **Nit**
This is triggering many times right? Would be more interesting to add a guard for phase or label change only which should be the important part.

pkg/rancher-ai-ui/utils/format.ts
```diff
@@ -137,6 +131,25 @@ export function formatAgentMetadata(data: string, agents: Agent[]): AgentMetadat
...
+
+    return {
+      agent,
+      query: parsed.query
+    };
+  } catch (err) {
+    error('Failed to parse sub-agent processing metadata:', err);
+  }
+
+  return null;
```
marcelofukumoto, line 154: **Nit**
You could make formatChatMetadata, formatAgentMetadata and formatSubAgentProcessingMetadata, use the same functions since mostly it is a dedup.

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
```diff
@@ -399,63 +366,78 @@ export function useChatMessageComposable(
...
         }
-      }
-      break;
+      });
+    } catch (err) {
+      warn('Failed to fetch UI tools calls for the Welcome message:', err);
+    } finally {
+      currentMsg.value.completed = true;
+
+      setProcessingState({ phase: MessagePhase.Idle });
```
marcelofukumoto: **Suggestion/Question**
Maybe a concern here.
This is a async request using the currentMsg. The currentMsg ideally can change to be the streaming message no?
Just thinking that would be nice to have a guard to make sure the currentMsg is still the one we were expecting.

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
```diff
@@ -464,13 +446,29 @@ export function useChatMessageComposable(
 
       break;
     default:
-      setPhase(MessagePhase.GeneratingResponse);
+      setProcessingState({ phase: MessagePhase.GeneratingResponse });
```
marcelofukumoto, line 449: **Suggestion**
setProcessingState here is not a good idea because it will set it again on some conditions. So it would be better to add after all the others were not added because of a break.

marcelofukumoto (commented):
I did some tests but I am not sure because it doesn't feel I triggered any subAgent.
As a regression it seems to be working fine.
Checking the code I have some simple comments.

The flow is quite complex so if you want to discuss on one of the points that is great.
Overall seems are only optimisations.
Some problems could happen for edge cases specially for racing conditions.

cypress/e2e/po/chat.po.ts
torchiaf: We can't use any elements visibility to wait before scrolling up.
The problem here is that we have to wait until  the scroll composable finishes to calculate panel height -> the scroll button visibility.
If the scroll happens immediately after the last message completes, the scroll button doesn't pop up in the Chat.

I fixed the `scrollTop` function to use the native `scrollTo` so now it runs smoothly, but still needs a 100ms duration.

pkg/rancher-ai-ui/composables/__tests__/useChatMessageComposable.test.ts
torchiaf: This is a typo - removed.

pkg/rancher-ai-ui/l10n/en-us.yaml
torchiaf: "sub-agents" is a bit misleading.

Previous architecture (called multi-agents): The assistant selects ALWAYS 1 agent to perform the request
Current architecture (called sub-agents): The assistant creates a "supervisor" agent -> the supervisor selects {1..N} agents to perform the request.

So (s) is to highlight that it could select 1 or more Agents, in adaptive mode

pkg/rancher-ai-ui/components/Processing.vue
torchiaf: Done

pkg/rancher-ai-ui/store/chat.ts
torchiaf: I'd leave it as is, the processing label in the HTML is responsive only when the `processingState.phase` changes. Also, they should be handled as a single payload, otherwise we risk to save a phase with an incorrect  label.

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
torchiaf: True - we don't need to use `currentMsg` here, it's safer to just use a local variable. Fixed.

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
torchiaf: If we move it at the end of default case, it means we would call `setProcessingState` before each break.
I think keep it here is pretty safe, even because the phase is debounced, so rapid updates are not visible.

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
torchiaf: Yeah, this could be better - I pushed this commit for cleaning-up 2e69014d7bc37a69aa9a31db2b4bb3140927245e

pkg/rancher-ai-ui/utils/format.ts
torchiaf: I changed the formatSubAgentProcessingMetadata [here](2e69014d7bc37a69aa9a31db2b4bb3140927245e) but it complicates things if we want to make a generic function for all of those - let's leave them as is.

torchiaf:
Thanks for the review @marcelofukumoto , I pushed the requested changes - please take a look.

pkg/rancher-ai-ui/l10n/en-us.yaml
marcelofukumoto: Hmmm so the subagents created are not subagents right? They are other agents?
From my perspective, subagents were agents created by the agent to fulfil specific tasks. 
So the concept for me is slightly different. Would be nice to discuss on this to create a nice concept of definitions.

marcelofukumoto:
I checked the changes all good. I will just do a final test now.

marcelofukumoto:
I mostly tested on a regression perspective. I still couldn't trigger the sub agents to activate.
@torchiaf could you provide a prompt that would trigger it? And is the env you shared with me with the new backend changes applied?

marcelofukumoto:
One final question @torchiaf 
The Fleet Agent is called Fleet, should we call it Continuous Delivery Agent instead? 
Maybe not for now, or at all, but just questioning to make sure it is ok.

marcelofukumoto (approved):
I was able to test, and was able to interact with the sub agents. 
I tested the ability to show the message to change to Manual.
And I tested whenever you select an option different than the Adaptive one, it only uses one agent.

Already approved the code.
Overall, LGTM.

torchiaf:
> One final question @torchiaf The Fleet Agent is called Fleet, should we call it Continuous Delivery Agent instead? Maybe not for now, or at all, but just questioning to make sure it is ok.

@marcelofukumoto Yes, definitely - opening a new ticket: https://github.com/rancher/rancher-ai-ui/issues/245

PR #243, opened 2026-06-15
Fix related-resource buttons

### LINKED ISSUE #241: Support SUSE virtualization Related Resources
I'm implementing Harvester MCP, and when I ask questions that return Harvester resources, the "Related Resources" button redirects me to the wrong URL.

For example, here is a virtual machine button "View VirtualMachine: Ubuntu-0"

Image

It redirects me to:

`https://<rancher-url>/dashboard/c/c-m-8nvd5g8s/explorer/kubevirt.io.virtualmachine/ubuntu-0?as=yaml`

And the page can't show because the URL lacks a **namesapce**. If I change the URL to the following URL, the virtual machine resource is displayed. :

`https://<rancher-url>/dashboard/c/c-m-8nvd5g8s/explorer/kubevirt.io.virtualmachine/default/ubuntu-0?as=yaml`

**Note** I tried my best to compare other MCP servers in the rancher-ai-mcp, but can't figure out if the responder should carry some hints to indicate the resource is namespaced.  @torchiaf suggest that I open a ticket to track.

## Harvester dashboard integration

The above redirects me to the page inside the explorer. For Harvester resources, it's better to redirect to the Harvester dashboard like:

`https://<rancher-url>/dashboard/harvester/c/c-m-8nvd5g8s/kubevirt.io.virtualmachine/default/ubuntu-0`

Please advise if I need to change anything to support this.

torchiaf:
Fixes https://github.com/rancher/rancher-ai-ui/issues/241

### Overview

The Rancher dashboard is designed to generate details page's routes based on the current {cluster, product, store, resource_type}.
Resource buttons are currently based on this pattern, but this doesn't work on Liz, because the product or the cluster in which the user is operating may differ from those targeted by the buttons.

For example, when a user first navigates to the Deployments page in local cluster and Liz returns a resource button pointing to a Harvester virtual machine, the `detailLocation` value is incorrect because the resource model and inStore value belong to the `explorer` product.

We need to generate the `detailLocation` so that it doesn't depend on the current router location and the product in use.

The rule must be: "calculate the resource's 'owner' product and redirect to that product and cluster, regardless of the location at the time of the click".

- Native types (Pods, Deployments, etc...): We want to handle those resources in the  explorer
- CRDs: we identify the product handling that CRD. For instance, `kubevirt.io.virtualmachine` are handled in Harvester
- Fleet types: we check if a resource is a fleet resource - in this case we are free to statically do the check since Fleet is a built-in product

### Other improvements

Given the current changes, when the resource is from a downstream cluster, we have to load the schema and the resource itself via `/k8s/clusters/` APIs.
In some scenarios, like re-opening old chats, those API calls are sent all at the same time.

- In order to avoid performance issues, we want to fetch the data ONLY when the button is actually visible in the chat. Once a resource is loaded, we keep it in memory to avoid unnecessary new API calls.

### Technical changes

- Decouple `Action` component in simple RouteButtons and ResourceButtons
- Implement visibility observer
- Implement `resource-context` functions to generate the correct model for the resources.
- Adding e2e tests
  - Note that some tests are todo, we want to complete them once https://github.com/rancher/rancher-ai-ui/issues/250 is done
  
### How to test

Testing all the different scenarios using a real LLM is complex - generating buttons for certain schemas, such as those from Harvester or Elemental, would require a dedicated MCP.
One of the approaches below are fine anyway:

- Mock the mcp responses using the llm-mock API, see the e2e tests
- Hardcode the mcp responses in the message composable: https://github.com/rancher/rancher-ai-ui/commit/87748853201187f624be212b971eed3de0309a85

### Screenshots

Below a quick demo; it's now possible to navigate from a product/cluster to others without breaking the navigation (do not consider environment's slowness, it's for other reasons).

https://github.com/user-attachments/assets/1905b688-86ba-4c4a-b022-04096c7e28fc


pkg/rancher-ai-ui/components/message/resource-buttons/index.vue
```diff
@@ -42,6 +48,34 @@ const remaining = computed(() => {
...
+
+/**
+ * Observe the visibility of the resource buttons container and update isVisible.
+ * This is used to determine when to load the schema and resource for the buttons
+ * to avoid unnecessary API calls when the buttons are not visible in the viewport.
+ */
+function observeButtonsVisibility() {
+  visibilityObserver.value = new IntersectionObserver((entries) => {
+    entries.forEach((entry) => {
+      isVisible.value = !!entry.isIntersecting;
```
torchiaf, line 61: This return `true` when the buttons becomes visible, for instance when scrolling up to a previous message

pkg/rancher-ai-ui/components/message/resource-buttons/resource-context.ts
```diff
@@ -0,0 +1,320 @@
...
+      type:     normalizedType,
+      id:       normalizedId,
+      metadata: {
+        name,
+        namespace: normalizedNamespace,
+      }
+    };
+
+    // Get the correct store context
+    const storeCtx = store.getters[`${ inStore }/classify`];
```
torchiaf, line 259: Core functionality: we use `classify` getter and selected product to generate the correct `detailLocation`.
The `inStore` is based on selected product instead of the current store context.

pkg/rancher-ai-ui/components/message/resource-buttons/resource-context.ts
```diff
@@ -0,0 +1,320 @@
...
+  }
+
+  for (const product of activeProducts) {
+    if (product.inExplorer) {
+      continue;
+    }
+
+    const allTypes = store.getters['type-map/allTypes'](product.name)?.['all'] || {};
+
+    if (allTypes[resourceType]) {
```
torchiaf, line 214: Core funtionality: we check if the schema is included in those handled by the product

pkg/rancher-ai-ui/components/message/resource-buttons/index.vue
```diff
@@ -21,6 +23,10 @@ const props = defineProps({
   },
 });
 
+const resourceButtonsRef = ref<HTMLDivElement | null>(null);
```
rak-phillip: nit: we can utilize `useTemplateRef` to more clearly express that this ref is meant for DOM access[^1]. It's not required and the outcome will be the same with both approaches.

[^1]: https://vuejs.org/guide/essentials/template-refs.html#template-refs

pkg/rancher-ai-ui/components/message/resource-buttons/ResourceButton.vue
```diff
@@ -0,0 +1,224 @@
...
+        type: normalizedType,
+        id,
+      });
+    } else if (cluster === 'local') {
+      await store.dispatch(`${ inStore.value }/find`, {
+        cluster,
+        type: normalizedType,
+        id,
+      });
+    } else {
```
rak-phillip: nit: both of the statements withing these clauses look identical. Consider collapsing into a single clause

pkg/rancher-ai-ui/components/message/resource-buttons/resource-context.ts
```diff
@@ -0,0 +1,320 @@
...
+const MANAGEMENT_CLOUD_CREDENTIAL = 'management.cattle.io.cloudcredential';
+
+/**
+ * Check if the given schema represents a Custom Resource Definition (CRD)
+ *
+ * @param schema - The schema object to check
+ * @returns True if the schema represents a Custom Resource Definition (CRD), false otherwise
+ */
+export function isCRD(schema: Schema | null): boolean {
+  return schema?.attributes.crd === true || schema?.attributes.crd === 'true';
```
rak-phillip: Should we guard against situations where `attributes` could be `null` or `undefined`?

pkg/rancher-ai-ui/components/message/resource-buttons/ResourceButton.vue
```diff
@@ -0,0 +1,224 @@
...
+</script>
+
+<template>
+  <div ref="resourceButtonRef">
+    <RcButton
+      small
+      variant="secondary"
+      :disabled="!resource"
+      :data-testid="`rancher-ai-ui-chat-message-resource-button-${ props.value?.resource?.name }`"
+      @click="goTo"
```
rak-phillip, line 203: Consider if the `to` prop will be a better fit here[^1]. I think that it would work better for accessibility, but `goTo()` would require a bit of refactoring. 

[^1]: https://rancher.github.io/storybook/?path=/docs/stories-components-rcbutton--docs#as-router-link

rak-phillip (approved):
LGTM - I have a few optional comments to consider before merging. Otherwise, the changes are clean and well-formed.

pkg/rancher-ai-ui/components/message/resource-buttons/index.vue
torchiaf: Fixed

pkg/rancher-ai-ui/components/message/resource-buttons/ResourceButton.vue
torchiaf: I'd say let's keep this simple - we don't need a `<a>` link here

PR #261, opened 2026-07-08
Fix confirmation request 

### LINKED ISSUE #262: [Bug] Specific Agent mode requires double confirmation due to missing agent context in response
When interacting with the AI Assistant using a Specific Agent mode (e.g., Rancher), users are incorrectly prompted twice to confirm the same action before it executes.

This occurs because the UI does not include the agent parameter in the confirmation response payload over the WebSocket. Without this context, the backend supervisor is invoked again and attempts to evaluate/create the resource from scratch, resulting in a secondary confirmation prompt. This friction does not occur when using the Adaptive Agents Selection.

Environment Details:
```
# kubectl get apps -n cattle-ui-plugin-system
NAME            CHART           VERSION         RELEASE NAME    RELEASE VERSION   STATUS
rancher-ai-ui   rancher-ai-ui   1.1.0-dev-5     rancher-ai-ui   1                 deployed

# kubectl get deployments -n cattle-ai-agent-system -o custom-columns="DEPLOYMENT:.metadata.name,IMAGE(S):.spec.template.spec.containers[*].image"
DEPLOYMENT           IMAGE(S)
rancher-ai-agent     stgregistry.suse.com/rancher/rancher-ai-agent:v1.1.0-alpha.8
rancher-mcp-server   stgregistry.suse.com/rancher/rancher-ai-mcp:v1.1.0-alpha.7
```
**LLM: gemini-2.5-pro**

Steps to Reproduce:

1. Prompt the agent to create a resource (e.g., `create nginx deployment in local cluster default ns`).
2. Wait for the assistant to generate the payload and ask for confirmation.
3. Click Confirm.
4. Observe the UI/Network tab: The assistant repeats the exact same payload and confirmation prompt.
5. Click Confirm a second time to actually execute the action.

https://github.com/user-attachments/assets/c616545f-5851-4904-a862-5fffd8fa3a73

Image

torchiaf:
Fixes https://github.com/rancher/rancher-ai-ui/issues/262

- We are sending now the agent name in the confirmation payload as required in the backend

### Notes

This is a regression from Subagents `feature`, the target release is 1.1.0 - we don't need a backport for this.

### DIFF
--- a/cypress/e2e/po/message.po.ts
+++ b/cypress/e2e/po/message.po.ts
@@ -51,7 +51,7 @@ class RawMessagePo extends ComponentPo {
   }
 
   confirmationStatus() {
-    return this.self().get('[data-testid="rancher-ai-ui-chat-message-confirmation-status"]');
+    return this.self().get('[data-testid^="rancher-ai-ui-chat-message-confirmation-status-"]');
   }
 
   isCompleted() {
@@ -59,11 +59,11 @@ class RawMessagePo extends ComponentPo {
   }
 
   isConfirmed(args = { withLabel: 'Confirmed' }) {
-    return this.confirmationStatus().get('.status-confirmed').should('contain.text', args.withLabel);
+    return this.self().get('[data-testid="rancher-ai-ui-chat-message-confirmation-status-confirmed"]').should('contain.text', args.withLabel);
   }
 
   isCanceled(args = { withLabel: 'Canceled' }) {
-    return this.confirmationStatus().get('.status-canceled').should('contain.text', args.withLabel);
+    return this.self().get('[data-testid="rancher-ai-ui-chat-message-confirmation-status-canceled"]').should('contain.text', args.withLabel);
   }
 
   thinkingLabel() {
--- a/cypress/e2e/po/message.po.ts
+++ b/cypress/e2e/po/message.po.ts
@@ -50,16 +50,20 @@ class RawMessagePo extends ComponentPo {
     return this.self().get('[data-testid="rancher-ai-ui-chat-message-confirmation-cancel-button"]');
   }
 
+  confirmationStatus() {
+    return this.self().get('[data-testid="rancher-ai-ui-chat-message-confirmation-status"]');
+  }
+
   isCompleted() {
     return this.self().get(`[data-teststatus="rancher-ai-ui-chat-message-status-${ this.id }-completed"]`).should('exist');
   }
 
-  isConfirmed() {
-    return this.self().get('[data-testid="rancher-ai-ui-chat-message-confirmation-confirmed"]').should('exist');
+  isConfirmed(args = { withLabel: 'Confirmed' }) {
+    return this.confirmationStatus().get('.status-confirmed').should('contain.text', args.withLabel);
   }
 
-  isCanceled() {
-    return this.self().get('[data-testid="rancher-ai-ui-chat-message-confirmation-canceled"]').should('exist');
+  isCanceled(args = { withLabel: 'Canceled' }) {
+    return this.confirmationStatus().get('.status-canceled').should('contain.text', args.withLabel);
   }
 
   thinkingLabel() {
--- a/cypress/e2e/tests/features/multi-agent/message.spec.ts
+++ b/cypress/e2e/tests/features/multi-agent/message.spec.ts
@@ -87,14 +87,20 @@ describe('Multi Agent Messages', () => {
           name: 'createKubernetesResource',
           args: {
             kind:      'Pod',
-            name:      'my-pod',
+            name:      'my-pod-2',
             resource:  {
               apiVersion: 'v1',
               kind:       'Pod',
               metadata:   {
-                name:      'my-pod',
+                name:      'my-pod-2',
                 namespace: 'default'
               },
+              spec: {
+                containers: [{
+                  name:  'my-container',
+                  image: 'nginx'
+                }]
+              }
             },
             cluster:   'local',
             namespace: 'default'
@@ -112,9 +118,10 @@ describe('Multi Agent Messages', () => {
     confirmationRequestMessage.isCompleted();
     confirmationRequestMessage.containsText('Are you sure you want to proceed with this action?');
     confirmationRequestMessage.confirmButton().click();
-    confirmationRequestMessage.isConfirmed();
-    confirmationRequestMessage.containsText('Confirmed');
+
     confirmationRequestMessage.agentSelectionLabel().contains('Adaptive Agent(s) Selection');
+    confirmationRequestMessage.resourceButton({ name: 'my-pod-2' }).should('exist');
+    confirmationRequestMessage.isConfirmed();
 
     let resultMessage = chat.getMessage(6);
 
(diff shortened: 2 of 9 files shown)


cypress/e2e/po/message.po.ts
```diff
@@ -50,16 +50,20 @@ class RawMessagePo extends ComponentPo {
     return this.self().get('[data-testid="rancher-ai-ui-chat-message-confirmation-cancel-button"]');
   }
 
+  confirmationStatus() {
```
torchiaf, line 53: a bit of refactoring here

cypress/e2e/tests/features/multi-agent/message.spec.ts
```diff
@@ -87,14 +87,20 @@ describe('Multi Agent Messages', () => {
...
             resource:  {
               apiVersion: 'v1',
               kind:       'Pod',
               metadata:   {
-                name:      'my-pod',
+                name:      'my-pod-2',
                 namespace: 'default'
               },
+              spec: {
+                containers: [{
```
torchiaf, line 99: We are now passing a complete payload and actually create the resource in the cluster. Same in other test specs

cypress/e2e/tests/features/multi-agent/message.spec.ts
```diff
@@ -112,9 +118,10 @@ describe('Multi Agent Messages', () => {
     confirmationRequestMessage.isCompleted();
     confirmationRequestMessage.containsText('Are you sure you want to proceed with this action?');
     confirmationRequestMessage.confirmButton().click();
-    confirmationRequestMessage.isConfirmed();
-    confirmationRequestMessage.containsText('Confirmed');
+
     confirmationRequestMessage.agentSelectionLabel().contains('Adaptive Agent(s) Selection');
+    confirmationRequestMessage.resourceButton({ name: 'my-pod-2' }).should('exist');
```
torchiaf, line 123: Checking if the resource is really created. Same in other tests

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
```diff
@@ -160,6 +160,7 @@ export function useChatMessageComposable(
   function confirmMessage({ message, result }: { message: Message; result: boolean }, ws: WebSocket) {
     wsSend(ws, formatWSInputMessage({
       prompt: result ? ConfirmationResponse.Yes : ConfirmationResponse.No,
+      agent:  message.agentMetadata?.agent?.name || undefined, // This is required when Agent seleciton is manual
```
torchiaf: Adding the agent name here - this is a backend requirements with subagents architecture

marcelofukumoto:
Tried with the main version, doesn't even trigger properly.

Image
image

It feels it is having some problems, which feels it is the same cause which would be not having the proper agent selected to do the action.

With the fix:
image
image

This one feels that works a lot better.

----

Just adding as information
The Rancher Agent sometimes is not chosen for this action and the Adaptive didnt want to follow up with the deployment creation.

pkg/rancher-ai-ui/composables/useChatMessageComposable.ts
marcelofukumoto: Typo on seleciton -> selection.

cypress/e2e/po/message.po.ts
```diff
@@ -50,16 +50,20 @@ class RawMessagePo extends ComponentPo {
...
+  }
+
   isCompleted() {
     return this.self().get(`[data-teststatus="rancher-ai-ui-chat-message-status-${ this.id }-completed"]`).should('exist');
   }
 
-  isConfirmed() {
-    return this.self().get('[data-testid="rancher-ai-ui-chat-message-confirmation-confirmed"]').should('exist');
+  isConfirmed(args = { withLabel: 'Confirmed' }) {
+    return this.confirmationStatus().get('.status-confirmed').should('contain.text', args.withLabel);
```
marcelofukumoto: I think we should keep using the data-testid.

pkg/rancher-ai-ui/components/message/Confirmation.vue
marcelofukumoto: We should make SystemRequest and Confirmation a little more similar, you added the class on the SystemRequest through a variable so we could replicate it here and add the similar testId to both of them.

marcelofukumoto (commented):
Seems ok after my test.
But I have only one question additionally:

- When we have a confirmation message, ideally we already have an AGENT setup. So shouldn't we send to the Adaptive mode the proper AGENT that is already selected?

It feels like it is one AGENT that is making the confirmation question, so we need to send it back to the same one.

torchiaf:
> Seems ok after my test. But I have only one question additionally:
> 
> * When we have a confirmation message, ideally we already have an AGENT setup. So shouldn't we send to the Adaptive mode the proper AGENT that is already selected?
> 
> It feels like it is one AGENT that is making the confirmation question, so we need to send it back to the same one.

@marcelofukumoto We can't do that because the selected agent in Adaptive mode is not visible to the UI - in that scenario the supervisor (the main agent in the pod agent) is responsible to select the agent(s) so it already have such info.
In Manual mode, it bypasses the supervisor, so we have to send the agent from UI

torchiaf:
@marcelofukumoto I think I addressed all requests, please take a look

marcelofukumoto:
> > Seems ok after my test. But I have only one question additionally:
> > 
> > * When we have a confirmation message, ideally we already have an AGENT setup. So shouldn't we send to the Adaptive mode the proper AGENT that is already selected?
> > 
> > It feels like it is one AGENT that is making the confirmation question, so we need to send it back to the same one.
> 
> @marcelofukumoto We can't do that because the selected agent in Adaptive mode is not visible to the UI - in that scenario the supervisor (the main agent in the pod agent) is responsible to select the agent(s) so it already have such info. In Manual mode, it bypasses the supervisor, so we have to send the agent from UI

The problem with the Adaptive mode if that is the case and if there is a problem needs to be treated on the backend?
Maybe something to recheck later if can be easily replicated.

pkg/rancher-ai-ui/components/message/Confirmation.vue
```diff
@@ -126,31 +126,15 @@ const confirmationText = computed(() => {
...
-          <i class="icon icon-close canceled" />
-          <p>
-            {{ t('ai.confirmation.canceled') }}
-          </p>
-        </div>
-      </template>
+      <i :class="['icon', `icon-${ props.message.confirmation.status === ConfirmationStatus.Confirmed ? 'checkmark' : 'close'}` ]" />
+      <p>
+        {{ t(`ai.confirmation.${ props.message.confirmation.status }`) }}
+      </p>
```
marcelofukumoto: Most a NIT
It would be interesting if we have an object mapping those cases to be easy to read.

```
const STATUS_PRESENTATION: Partial<Record<ConfirmationStatus, { icon: string, label: string }>> = {
  [ConfirmationStatus.Confirmed]: {
    icon:  'icon-checkmark',
    label: 'ai.confirmation.confirmed'
  },
  [ConfirmationStatus.Canceled]: {
    icon:  'icon-close',
    label: 'ai.confirmation.canceled'
  },
};
```

That makes the translation tags fully grepable, more clear what are the cases and the icons for each case.

marcelofukumoto (approved):
That is just a NIT, a small suggestion on how I thought it would be interesting to have this structure.
I am not a fan of dynamic i18n tags and not a fan of ternary as well.

But only NIT in my opinion.

LGTM, I tested again and works fine.

pkg/rancher-ai-ui/components/message/Confirmation.vue
torchiaf: Yeah, that makes sense, I pushed the change, thanks.

marcelofukumoto (approved):
Thank you for applying my last Nit
LGTM

PR #263, opened 2026-07-10
Support Model x Agent settings

### LINKED ISSUE #257: Run Agent's workflows

torchiaf:
Fixes https://github.com/rancher/rancher-ai-ui/issues/257

We want to add the possibility to assign a custom model to each Agent.

  - The default model is used by default for each legacy & new Agent.
  - When no models are available, the model selector in the Agent configs section is disabled
  - When the list of available models changes and the new list does not include the custom model, all custom models are reset to the default model. This can happen when:
    - The user selects a new provider, in this case the list of available models is updated with the models from the new provider.
    - The list of the models is populated by third-party provider's endpoints and some of the initial parameters change, giving a new list of models. For instance, when Bedrock is selected the list of models can change if the user updates the `region`.
  - When there is a model reset, we display a `growl` message to inform the user that his action was reflected to the Agent configurations.
  - When the user moves back to the initial configuration stored in the cluster (same provider, same models) ~~the custom models assigned to the agents is restored~~ we keep the default model and the llmModelEnable is set to false

### Screenshots

- Agent configs - Model's selector

image

image

- Model's reset warning message

image

 - Model selection in action

### DIFF
--- a/pkg/rancher-ai-ui/pages/settings/sections/ai-agent-configs/index.vue
+++ b/pkg/rancher-ai-ui/pages/settings/sections/ai-agent-configs/index.vue
@@ -605,10 +605,6 @@ watch(validationErrors, (errors) => {
         <div class="form-values-row">
           <h3 class="m-0">
             {{ t('aiConfig.form.section.aiAgent.sections.llmModel.title') }}
-            <span
-              class="required"
-              aria-hidden="true"
-            >*</span>
             <i
               v-clean-tooltip="t('aiConfig.form.section.aiAgent.sections.llmModel.tooltip')"
               class="icon icon-info tooltip-icon"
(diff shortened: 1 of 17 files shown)


marcelofukumoto:
When I change the model
image

It shows the one I selected not the default one.
Seems a little misguiding.

I think the moment I click on the Same as AI provider, it should show the actual default.
Not sure if exactly on the select box, but it needs to be somewhere.

pkg/rancher-ai-ui/pages/settings/sections/ai-agent-configs/index.vue
```diff
@@ -558,6 +602,41 @@ watch(validationErrors, (errors) => {
           </div>
         </div>
 
+        <div class="form-values-row">
+          <h3 class="m-0">
+            {{ t('aiConfig.form.section.aiAgent.sections.llmModel.title') }}
+            <span
+              class="required"
```
marcelofukumoto: It is not required, is it ? Not sure. Weird since it is a checkbox and... I don't know. A little misguiding.

marcelofukumoto (commented):
**Question**
Check if we actually need the *.

torchiaf:
@marcelofukumoto the last commit fixes the watcher firing multiple times even when the models didn't change - now the activeChatbot and model updates emit the models list but from outside the watcher; the watcher is now working only for models list changes.

marcelofukumoto (approved):
Thank you for the many minutious changes!
It was a fun ride!
LGTM

PR #266, opened 2026-07-21
E2E Update to use HELM instead of DOCKER

marcelofukumoto:
Changes pulled from the extension compatibility test.
I executed the run 5 times:
https://github.com/marcelofukumoto/rancher-ai-ui/actions/workflows/test.yaml

https://github.com/marcelofukumoto/rancher-ai-ui/actions/runs/29768357154 [NG]
https://github.com/marcelofukumoto/rancher-ai-ui/actions/runs/29811775058 [OK]
https://github.com/marcelofukumoto/rancher-ai-ui/actions/runs/29765254545 [NG]
https://github.com/marcelofukumoto/rancher-ai-ui/actions/runs/29762699557 [OK]
https://github.com/marcelofukumoto/rancher-ai-ui/actions/runs/29752521141 [OK]
https://github.com/marcelofukumoto/rancher-ai-ui/actions/runs/29750496885 [OK]

For the NG ones seems they are usual flaky tests but you can check as well.

The process was:
- First setup the entire process without thinking it would be added on E2E to check how that needs to be fully setup using HELM.
- Than check the DOCKER setup and compare
- Introduce the E2E environment and the setup done on the RANCHER/DASHBOARD by Ricahrd.
- Check for differences related to extension and the setup for Rancher-AI-UI to work
- After done, run the execution and see for "flakiness"

Later we found out the current flakiness on the E2E of rancher-ai-ui.
So changed the approach to check the current flakiness of the MAIN branch.
Accept some flakiness but fix others to make sure the setup would work better.

This is the final result and the description done by the AI of the operation.

---

# AI GENERATED DESCRIPTION
# Migrate e2e from Docker all-in-one to Helm-on-k3s (same-origin, serve-pkgs)

Mirrors the approach in [rancher/dashboard#10897]: run the e2e suite against a **real Kubernetes (k3s) cluster with Rancher installed via Helm**, and load the `rancher-ai-ui` extension into Rancher's **own dashboard** from a local catalog server — instead of the previous Docker all-in-one image + `yarn dev` proxy.

## Why

The old setup ran Rancher as a single `rancher/rancher:head` container and served the UI from a separate `yarn dev` process that proxied API calls back to it. That works, but it doesn't reflect how the extension actually ships (developer-loaded into Rancher, same-origin) and pins us to the all-in-one image. This moves the tests onto a topology that matches production loading and a standard Helm-on-k3s install.

---

## Before — Docker all-in-one + `yarn dev`

```
┌─ GitHub runner ─────────────────────────────────────────────┐
│                                                             │
│  docker run rancher/rancher:head  (--privileged, :80/:443)  │
│     └─ bundled k3s + Rancher (one tuned container)          │
│                                                             │
│  yarn dev  ──►  dev server (https://localhost:8005)         │
│     └─ serves the extension, PROXIES /v1 /v3 /api ──► Rancher │
│                                                             │
│  Cypress  ──►  baseUrl = https://localhost:8005             │
└─────────────────────────────────────────────────────────────┘
```

- **Rancher:** `docker run -d --privileged -p 80:80 -p 443:443 rancher/rancher:head` — one container bundling its own k3s.
- **UI:** `yarn dev` dev server; Cypress `baseUrl` points at the dev server root.
- **API calls:** relative (`/v1`, `/v3`, `/api`) — the dev server proxied them to Rancher, so they "just worked."
- **Workflow:** `Start Rancher (docker)` → `Deploy AI charts` → `Start dev UI (yarn dev)` → `wait-on dev UI` → `Cypress`.

## After — Helm-on-k3s + serve-pkgs (same-origin developer-load)

```
┌─ GitHub runner ──────────────────────────────────────────────────┐
│                                                                  │
│  k3s (v1.36.1+k3s1)                                               │
│   ├─ cert-manager                                                │
│   ├─ Rancher (helm install, hostname 127.0.0.1.sslip.io)         │
│   │    └─ serves its OWN dashboard at /dashboard  (Traefik/TLS)  │
│   ├─ rancher-ai-agent + llm-mock + mcp-server (helm)             │
│   └─ UIPlugin CR ──► developer-load extension from catalog       │
│                                                                  │
│  serve-pkgs catalog server (http://127.0.0.1:8080)               │
│     └─ built extension bundle, registered via the UIPlugin        │
│                                                                  │
│  Cypress  ──►  baseUrl = https://127.0.0.1.sslip.io/dashboard    │
│                (SAME ORIGIN as Rancher — no dev proxy)           │
└──────────────────────────────────────────────────────────────────┘
```

- **Cluster:** install k3s directly on the runner, wait for the node Ready. The k3s version and the Rancher chart channel are script parameters (default `latest`); CI pins the validated pair.
- **Rancher:** `cert-manager` (jetstack) + `helm install rancher` with `hostname=127.0.0.1.sslip.io`; wait for the `rancher` deployment rollout and for `/dashboard/` to answer.
- **Extension loading:** `yarn build-pkg` → **serve-pkgs** catalog server on `:8080` → register a **`UIPlugin` CR** (developer-load, `noCache`) so Rancher's own dashboard pulls the extension bundle. No `yarn dev`.
- **Same-origin:** Cypress `baseUrl = https://127.0.0.1.sslip.io/dashboard` — the UI and API share one origin, exactly like a real developer-load.
- **First-login:** the setup spec performs the UI first-login (bootstrap password, EULA, server-url) before the feature specs run.
- **Prime gate:** the extension is prime-only; when developer-loaded into community Rancher it self-gates off. We intercept `/rancherversion` in Cypress and set `RancherPrime = 'true'` (what the `yarn dev` PRIME proxy did) so the extension activates.
- **Workflow:** `Build extension` → `Start serve-pkgs` → `Start Rancher (k3s/Helm)` → `Deploy AI charts` → `Register extension (UIPlugin)` → `Cypress`.

### At a glance

| | Before (Docker) | After (Helm-on-k3s) |
|---|---|---|
| Rancher | `rancher/rancher:head` all-in-one container | Helm install on k3s (+ cert-manager, Traefik, TLS) |
| UI delivery | `yarn dev` dev server (separate origin) | Rancher's own dashboard, developer-loaded via `UIPlugin` CR |
| Extension source | `yarn dev` build | `serve-pkgs` catalog server + `yarn build-pkg` |
| Cypress `baseUrl` | `https://localhost:8005` (dev server root) | `https://127.0.0.1.sslip.io/dashboard` (same origin) |
| API request path | relative, proxied by dev server | absolute to server root (see below) |
| kubeconfig | n/a | direct k3s kubeconfig on disk |
| Install script | `install-rancher-docker.sh` | `install-rancher-helm.sh` |

Both install scripts are retained — the Docker one still backs `e2e-automation-spec-runner.yml` and local development. See [Review feedback](#review-feedback-4th-commit).

---

## API request routing (required by the new topology)

Because the UI is now served from Rancher at the **`/dashboard` subpath** and there is **no dev-server proxy**, relative API paths resolve against `baseUrl` and hit the dashboard SPA instead of the API:

```
cy.request('/v1/...')  ──►  https://…/dashboard/v1/...   ❌  returns dashboard HTML (200)
```

A small shared helper strips the `/dashboard` suffix so requests target the Rancher **server root**:

```ts
// cypress/support/utils/rancher-url.ts
export function rancherApiUrl(path: string): string {
  const base = (Cypress.config('baseUrl') || '').replace(/\/dashboard\/?$/, '');
  return `${ base }${ path }`;   // …/v1/...  ✅
}
```

- Applied across the `rancher-api`, `ai-service`, `llm-mock`, `chat-history` and `multi-agent` Cypress commands, and to the k8s **service-proxy** paths used to reach the agent / llm-mock (`/api/v1/namespaces/.../services/.../proxy`).
- **Backward-compatible:** in a `yarn dev` setup `baseUrl` has no `/dashboard` suffix, so the helper is a no-op.
- AI-service install/uninstall now resolves against the **direct k3s kubeconfig** (`KUBECONFIG_PATH`) rather than Rancher's `generateKubeconfig` action, so `helm uninstall` actually tears the agent down (the disconnection tests depend on that observable window).

Without this change the API-driven test setup (creating/deleting resources, cleaning chat history, controlling the llm-mock, agent-config CRUD) cannot run — every such request silently returns dashboard HTML.

---

## Test hardening

The k3s runner is slower and more contended than the tuned Docker appliance, which surfaced pre-existing timing brittleness in a handful of interactions. These specs are green on `main` (Docker); the fixes make them robust on the new environment:

- **Welcome greeting** — assert by text *presence* (the greeting is two `<span>`s and can be scrolled), not visibility.
- **History tooltip / per-item menu** — open with a synthetic `mouseenter`/`mouseover` (the proven `@rancher/cypress` `TooltipPo` pattern) instead of `realHover`/`realMouseUp`, which were ~50% flaky under load.
- **History `chatItems()` / `name()`** — scope with `find` (not `get`, which queries the whole document) and exclude the per-item menu button so menu elements aren't miscounted as chats.
- **Sliding-badge hooks** — scope hover/overlay to the first match (a second `bg-info` badge appears transiently while the resource state settles).
- **Resource-button lazy-load** — trigger the schema fetch with `scrollIntoView()` on the button, not a smooth panel scroll that can race layout after reopen and miss the `IntersectionObserver`.
- **Multi-agent selector** — wait for the agent config to reconcile *active* via the API before asserting the UI, instead of racing the controller.

Plus one **product-code** hardening (not test-only): `useAIServiceComposable`'s fire-and-forget `fetchDeployments` is now guarded, so a resource-cache teardown during an AI-service (re)install can't surface as an unhandled promise rejection. (The equivalent guard for `useAgentComposable` was dropped when rebasing — `main` has since landed its own, so that file is now untouched by this PR.)

---

## Known flaky test (pre-existing, not introduced here)

A 5-run baseline of **`main` (unchanged, Docker)** on this fork failed **2/5**, both on the same test:

> `features/multi-agent/chat.spec.ts` → "should not show the Adaptive Agent(s) Selection when only one agent is enabled and active"

The other 12 specs passed 5/5. This test depends on the agent controller reconciling an `AIAgentConfig` to `active` in time — a known upstream timing issue the test itself documents (`// TODO: … known issue with the controller that it does not update the status of an agent`). It flakes on Docker and k3s alike; this PR does not regress it, and the `waitForAgentConfigActive` wait added above reduces (but cannot eliminate) it. **Recommend tracking/quarantining separately upstream.**

---

## Review feedback (4th commit)

Addresses @torchiaf's review. Rebased onto current `main` first.

**Both Rancher install paths are kept.** `install-rancher.sh` was split into:

| Script | Used by | Topology |
|---|---|---|
| `install-rancher-docker.sh` | `e2e-automation-spec-runner.yml`, local development | all-in-one container + `yarn dev` |
| `install-rancher-helm.sh` | `test.yaml` (the PR gate) | Helm on k3s, developer-loaded extension |

`install-rancher-docker.sh` is restored **byte-for-byte** from `main` (git records it as a rename, zero content diff). This also fixes a break the earlier revision had introduced: `e2e-automation-spec-runner.yml` is a Docker + `yarn dev` flow that calls the same script, so replacing it in place would have had it install Helm-on-k3s at the wrong hostname.

**Versions are parameters now, defaulting to `latest`:**

```bash
# install-rancher-helm.sh [VERSION] [CATTLE_SERVER_URL] [CATTLE_BOOTSTRAP_PASSWORD] \
#                         [RANCHER_CHART_CHANNEL] [K3S_VERSION]
```

- `RANCHER_CHART_CHANNEL` replaces the hardcoded `release-2.15` in the chart repo URL.
- `K3S_VERSION` drives the k3s install. The installer is now fetched from `get.k3s.io` rather than pinned to a tag (`v1.35.3+k3s1`) that disagreed with the version actually being installed (`v1.36.1+k3s1`).
- `test.yaml` pins the validated combination (`release-2.15` / `v1.36.1+k3s1`) so CI stays on a known-good pair while the script default is `latest`. ⚠️ **The `latest` default path has not had a CI run yet** — worth one before merge.

**Inline workflow steps moved to scripts:** `start-extension-server.sh` (waits for the catalog to answer instead of `sleep 5`) and `register-extension.sh` (one `python3` call instead of three, and it validates the catalog before applying the CR).

**Cypress command fixes** in `resolveKubeconfig` — type the `cy.wrap` yield, and yield the path from the `cy.writeFile` chain. The latter was failing locally on the `generateKubeconfig` fallback path (CI never hit it, since CI sets `KUBECONFIG_PATH`).

**Controller-status workarounds** are now described in a single numbered `TODO` in the multi-agent chat spec, so the `systemPrompt` nudge and the `waitForAgentConfigActive` poll get removed together once the status update is fixed server-side. The poll is bounded and non-fatal, and logs the observed state plus the full status on timeout, so a stalled controller stays visible rather than being silently retried away.

**Also:** removed the temporary workflow triggers and the unused `CYPRESS_extension_server_url`; documented both install paths, the version parameters and `KUBECONFIG_PATH` in the README.

## Commits

| Commit | Scope |
|---|---|
| `e2e: migrate to Helm-on-k3s with same-origin serve-pkgs developer-load` | Migration infra: k3s + Helm Rancher, serve-pkgs, `UIPlugin` developer-load, prime-flag intercept, first-login |
| `e2e: route Cypress API/proxy requests to the Rancher server root` | The `rancherApiUrl` helper + direct-kubeconfig resolution (required by the same-origin topology) |
| `e2e: harden flaky test interactions and races` | The interaction/timing hardening above |
| `e2e: address review feedback on the Helm-on-k3s migration` | The review items above: install-script split, version parameters, extracted scripts, Cypress command fixes |

## Notes for reviewers

- The temporary `workflow_dispatch` + branch push triggers used to validate without a PR have been **removed**.
- CI runs on `ubuntu-latest` (4 vCPU / 16 GB); the single-node k3s cold-load is slow, so `TEST_TIMEOUT` is raised to `60000`.


cypress/support/commands/rancher-ai-service.ts
```diff
@@ -1,13 +1,28 @@
...
+ * (direct) kubeconfig the charts were installed with, so `helm uninstall` actually tears the
+ * agent down - which the disconnection tests depend on. Falls back to `generateKubeconfig`
+ * for environments that don't provide a direct kubeconfig.
  */
-Cypress.Commands.add('installRancherAIService', ( args: InstallRancherAIServiceArgs = { waitForAIServiceReady: true }) => {
+function resolveKubeconfig(): Cypress.Chainable<string> {
+  const directKubeconfig = Cypress.env('kubeconfig');
+
+  if (directKubeconfig) {
+    return cy.wrap(directKubeconfig, { log: false });
```
torchiaf: The return type is wrong: Type 'string' is not assignable to type 'JQuery<any>':

**Suggestion**:

```
return cy.wrap<string>(directKubeconfig, { log: false });
```

cypress.config.ts
```diff
@@ -1,12 +1,16 @@
...
 
 export default extendConfig({
-  env: {
+  // The extension catalog server (serve-pkgs) is plain http on :8080 while the dashboard is
+  // https; disabling web security lets the developer-loaded extension script load.
+  chromeWebSecurity: false,
+  env:               {
+    // Direct k3s kubeconfig path (set in CI). When present, the AI-service install/uninstall
+    // commands use it instead of Rancher's generateKubeconfig action - see resolveKubeconfig.
+    kubeconfig:   process.env.KUBECONFIG_PATH,
```
torchiaf, line 10: Could we mention the new var in the README file below ?

```
#### Launch Cypress

```bash
# Launch Cypress dashboard - interactive mode
TEST_SKIP=setup TEST_PASSWORD=${rancher-password} yarn cypress:open // Add the new var here

# Run Cypress tests in background
TEST_SKIP=setup TEST_PASSWORD=${rancher-password} yarn cypress:run
```

.github/scripts/install-rancher.sh
```diff
@@ -1,77 +1,156 @@
 #!/bin/bash
 
-# Helper script to run Rancher in a Docker container
```
torchiaf, line 3: Can we keep both Docker and Helm installation files? The Docker is sill useful for development

cypress/support/commands/rancher-ai-service.ts
```diff
@@ -19,24 +34,33 @@ Cypress.Commands.add('installRancherAIService', ( args: InstallRancherAIServiceA
 
       cy.writeFile(kubeconfig, resp.body.config);
```
torchiaf: This is throwing error when running the tests locally:

Image

**Suggested fix**:

We should change this:

```
  cy.writeFile(kubeconfig, resp.body.config);

  return kubeconfig;
```

To:

```
return cy.writeFile(kubeconfig, resp.body.config).then(() => kubeconfig);
```

.github/scripts/install-rancher.sh
```diff
@@ -1,77 +1,156 @@
...
+if [ -n "$1" ]; then VERSION=$1; fi
+if [ -n "$2" ]; then CATTLE_SERVER_URL="$2"; fi
+if [ -n "$3" ]; then CATTLE_BOOTSTRAP_PASSWORD="$3"; fi
+
+# ---------------------------------
+# ----------------------- Config (Rancher head / 2.15)
+# ---------------------------------
+
+KUBE_VERSION=${KUBE_VERSION:-v1.36.1+k3s1}
+RANCHER_HELM_REPO_URL=${RANCHER_HELM_REPO_URL:-https://charts.optimus.rancher.io/server-charts/release-2.15}
```
torchiaf: I think we should replace the hardcoded Rancher verion with an input param here, and use `latest` as default

.github/scripts/install-rancher.sh
```diff
@@ -1,77 +1,156 @@
...
-
-if [ -n "$2" ]; then
-  CATTLE_SERVER_URL="$2"
-fi
-
-if [ -n "$3" ]; then
-  CATTLE_BOOTSTRAP_PASSWORD="$3"
+echo ""
+echo "Installing k3s (${KUBE_VERSION}) ..."
+curl -sfL -o k3s-script https://raw.githubusercontent.com/k3s-io/k3s/v1.35.3%2Bk3s1/install.sh
```
torchiaf: Same for k3s version.

.github/workflows/test.yaml
```diff
@@ -27,32 +35,78 @@ jobs:
       - name: Setup Node env
         uses: ./.github/actions/setup
 
+      - name: Build extension package
+        run: yarn build-pkg rancher-ai-ui
+
+      - name: Start extension catalog server (serve-pkgs)
+        run: |
+          PORT=${{ env.EXTENSION_SERVER_PORT }} nohup yarn serve-pkgs > serve-pkgs.log 2>&1 &
```
torchiaf: Nit: we could move this one to a shell file

.github/workflows/test.yaml
```diff
@@ -27,32 +35,78 @@ jobs:
...
         env:
-          API: ${{ env.CATTLE_SERVER_URL }}
+          KUBECONFIG: ${{ env.KUBECONFIG_PATH }}
         run: |
-          nohup yarn dev > dev.log 2>&1 & echo $! > dev.pid
-
-      - name: Wait for dev UI to be ready
-        run: |
-          npx wait-on ${{ env.DEV_UI_URL }}
+          CATALOG=$(curl -s http://127.0.0.1:${{ env.EXTENSION_SERVER_PORT }}/)
```
torchiaf: Nit: we could move this one to a shell file

cypress/e2e/tests/features/multi-agent/chat.spec.ts
```diff
@@ -210,6 +210,10 @@ describe('Multi Agent Chat', () => {
         }
       });
 
+      // updateAgentConfig only PUTs the spec; wait for the controller to reconcile harvester back to
+      // active (2 active agents) before asserting the UI reflects it, instead of racing the reconcile.
+      cy.waitForAgentConfigActive(harvesterAgentConfig);
```
torchiaf: This is a workaround for the flakiness of the agent's status update right?

The polling inside the command will cause us to lose visibility of what is happening server-side (when it takes too much time to updating the Agent status).

I would include this line as part of the `TODO` change above so that we don't lose the context of the change. Once the status update is fixed server side, we can remove the TODO and waitForAgentConfigActive command

torchiaf (changes requested):
Overall this is a good improvement, we are fixing some Page Objects as well. Left some comments, please have a look.

cypress/support/commands/rancher-ai-service.ts
marcelofukumoto: Done

.github/scripts/install-rancher.sh
marcelofukumoto: Ok! Split into 2 now.

cypress.config.ts
marcelofukumoto: Added

cypress/support/commands/rancher-ai-service.ts
marcelofukumoto: Fixed by your suggestion

.github/scripts/install-rancher.sh
marcelofukumoto: Done, added as a param

.github/scripts/install-rancher.sh
marcelofukumoto: Done as well

.github/workflows/test.yaml
marcelofukumoto: Moved to register-extension.sh

.github/workflows/test.yaml
marcelofukumoto: Moved to start-extension-server.sh

cypress/e2e/tests/features/multi-agent/chat.spec.ts
marcelofukumoto: Yes, correct. Added the TODO tag

torchiaf (approved):
LGTM

Issue #245, opened 2026-06-17
Rename Fleet -> Continous Delivery


torchiaf:
Fleet Agent should be renamed with Continous Delivery in the UI

- Agent mode selection

<img width="313" height="231" alt="Image" src="https://github.com/user-attachments/assets/0ead7913-b88f-489b-af79-91a7e24713fd" />

- Agents settings

<img width="894" height="380" alt="Image" src="https://github.com/user-attachments/assets/aef5f7e5-0d4b-4b71-ba93-ac9e432a24b0" />

deepakpunia-suse:
### 🧪 Verification: SUCCESSFUL

Environment
- rancher-ai-ui: `1.1.0-rc.9`

<img width="598" height="305" alt="Image" src="https://github.com/user-attachments/assets/05c8db4a-5517-406b-8044-08f7d84da6e2" />

<img width="1441" height="530" alt="Image" src="https://github.com/user-attachments/assets/7135cb5c-c1bc-4703-b056-617ffd52bf40" />

PR #270, opened 2026-07-23
Update UI tools to support Continuous Delivery label

torchiaf:
Contributes to https://github.com/rancher/rancher-ai-ui/issues/245

This change updates the `explore` UI tool in order to generate "Continuous Delivery" routes instead of "Fleet"

### DIFF
--- a/cypress/e2e/tests/features/context.spec.ts
+++ b/cypress/e2e/tests/features/context.spec.ts
@@ -144,27 +144,4 @@ describe('Feature: context', () => {
 
     cy.get('[data-testid="rancher-ai-ui-chat-container"]').screenshot('context-test-7-selected-context-in-message');
   });
-
-  describe('disabled state', () => {
-    afterEach(() => {
-      cy.uninstallRancherAIService();
-      cy.installRancherAIService();
-    });
-
-    it('Test 8: Context panel is disabled when AI service is not active', () => {
-      cy.installRancherAIService({ waitForAIServiceReady: false });
-
-      const deploymentsListPage = new WorkloadsDeploymentsListPagePo('local', 'apps.deployment' as any);
-
-      deploymentsListPage.goTo();
-      deploymentsListPage.waitForPage();
-
-      cy.get('[data-testid="extension-header-action-ai.action.openChat"]').should('be.visible');
-      chat.open();
-
-      context.isDisabled();
-
-      cy.get('[data-testid="rancher-ai-ui-chat-container"]').screenshot('context-test-8-disabled-when-not-active');
-    });
-  });
 });
--- a/pkg/rancher-ai-ui/ui-tools.json
+++ b/pkg/rancher-ai-ui/ui-tools.json
@@ -182,8 +182,8 @@
             "clusters": {
               "description": "SCOPE: list or collection, this route shows the list of clusters in Rancher. Use this route when the context is about listing clusters or cluster management."
             },
-            "fleet": {
-              "description": "This route shows the Fleet dashboard. Use this route when the context is about managing or listing Fleet resources."
+            "continuous-delivery": {
+              "description": "This route shows the Continuous Delivery dashboard. Use this route when the context is about managing or listing Fleet resources."
             },
             "users": {
               "description": "This route shows the Users management page. Use this route when the context is about managing users, roles, or permissions."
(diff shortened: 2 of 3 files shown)


pkg/rancher-ai-ui/ui-tools.json
```diff
@@ -182,8 +182,8 @@
             "clusters": {
               "description": "SCOPE: list or collection, this route shows the list of clusters in Rancher. Use this route when the context is about listing clusters or cluster management."
             },
-            "fleet": {
-              "description": "This route shows the Fleet dashboard. Use this route when the context is about managing or listing Fleet resources."
+            "continuous-delivery": {
+              "description": "This route shows the Continuous Delivery dashboard. Use this route when the context is about managing or listing Fleet resources."
```
marcelofukumoto, line 186: **Question**
Should still be Fleet resources?

pkg/rancher-ai-ui/ui-tools.json
marcelofukumoto, line 118: **Question**
Do we update this value when we change things here?

marcelofukumoto: **Question**
Is this enum still correct?
Seems like it correlates with the previous one.

marcelofukumoto (commented):
Just some questions since this is mostly config.
So I mostly look for Fleet overall.
But the config on the ui-tools seems a blocker if it is wrong.

pkg/rancher-ai-ui/ui-tools.json
torchiaf: Ah Yes - fixed, thanks

pkg/rancher-ai-ui/ui-tools.json
torchiaf: Yes, that's because for Fleet resources the naming is mostly based on `fleet` keyword -> keeping "Fleet resources" in this prompt here helps the LLMs matching Fleet scope with Continuous Delivery label.
This was tested with Gemini 2.5 flash, which is faster but less smart than other models and it worked, so we can be pretty confident that it will work with other models.

pkg/rancher-ai-ui/ui-tools.json
torchiaf: When we update the tools definition, the UI will compare the current config with the new one built-in in the extension and will inform the user if needed:


https://github.com/user-attachments/assets/ad83905d-2fa3-41bb-8fab-185dd75f8fc7

marcelofukumoto: Ok... so do we need to update ? No?

torchiaf: Do you mean the revision? Not necessarily - updating the revision is needed when we want to reset a user value(enable/disable). In this case we can keep it as is

cypress/e2e/tests/features/context.spec.ts
```diff
@@ -144,27 +144,4 @@ describe('Feature: context', () => {
...
-      deploymentsListPage.waitForPage();
-
-      cy.get('[data-testid="extension-header-action-ai.action.openChat"]').should('be.visible');
-      chat.open();
-
-      context.isDisabled();
-
-      cy.get('[data-testid="rancher-ai-ui-chat-container"]').screenshot('context-test-8-disabled-when-not-active');
-    });
-  });
```
marcelofukumoto, line 169: Not sure why we have this deleted.

marcelofukumoto (approved):
Apart from the comment. LGTM

cypress/e2e/tests/features/context.spec.ts
torchiaf: This is not related to the current change - it's a duplicate of:
```
  describe('disabled state', () => {
    afterEach(() => {
      cy.uninstallRancherAIService();
      cy.installRancherAIService();
    });

    it('Test 7: Context panel is disabled while chat is not ready', () => {
      HomePagePo.goTo();

      cy.installRancherAIService({ waitForAIServiceReady: false });

      chat.open();

      chat.isNotReady();

      context.isDisabled();

      cy.get('[data-testid="rancher-ai-ui-chat-container"]').screenshot('context-selection-test-7-context-disabled');
    });
  });
```
in `cypress/e2e/tests/features/context-selection.spec.ts` file.

PR #272, opened 2026-07-24
Fix sliding badges

### LINKED ISSUE #244: Sliding badges various issues
We found some issues affecting the sliding badges:

- [x] 1. when a sliding badge is triggered, the sliding badge factory changes the default background color to make some colors 'solid', in order to prevent the sliding b. overlapping the parent. When a state change occurs (the badge background is updated) its state is corrupted by the style applied by the previous sliding badge.
- [x] 2. in any Details page -> resource tabs, we have some tables in each tab, in the DOM, but they are hidden by the active tab. However, the sliding badge factory creates the sliding badges even for hidden tables.
- [x] 3. in any Details page -> Header, the parent badge  is wrong, therefore the `destroy` can't find the right element to be removed. This is happening when using the easter egg trigger only.
- [x] 4. some background color are wrong because not-handled or wrongly handled by the factory.

torchiaf:
Fixes https://github.com/rancher/rancher-ai-ui/issues/244

The fixes are ordered by bugs reported in 244:

1. cce24e3c975c86e61cfcf32478ce322def4dff3c - we are now restoring the background color of the parent badge when the sliding badges are destroyed.

2. 325be42d35afdf901848e7d307668355037eb9ac - we are creating the sliding b. when the parent badge is visible only. Verified in the e2e test: `'Easter egg' -> 'It should show all sliding badges in a page'`

3. e099a546308388eb24ed99fb3ef93d3eb58015bb - fixed by using correct parent

4. 05ed1ff6cbca5d54aa5da5cfeb41236ddd55c53a - we are replacing the old mechanism based on class-id with a more robust approach, based on color transparency detection, which will dynamically work for all colors.
 
 
 ### Screenshots
 
1. before

https://github.com/user-attachments/assets/0568b8c3-7674-4445-a6f2-e4e108c692d6

1. after

https://github.com/user-attachments/assets/e5813e90-16bc-4efc-8d83-6aa4968920e6

3. before

https://github.com/user-attachments/assets/99c26eb2-fd53-44c6-970f-7c8a528d07f3

3. after

https://github.com/user-attachments/assets/83e1053d-5e23-4ac9-bc5a-a6685ebd9b89

4. before

https://github.com/user-attachments/assets/66b140eb-e2d0-49bc-a3c5-da757d75cad5

4. after

https://github.com/user-attachments/assets/b6be1b02-3f72-40e7-b368-47a59492fdaa

### DIFF
--- a/pkg/rancher-ai-ui/handlers/hooks/overlay/badge-sliding.ts
+++ b/pkg/rancher-ai-ui/handlers/hooks/overlay/badge-sliding.ts
@@ -13,6 +13,11 @@ const enum Theme {
   Dark = 'dark', // eslint-disable-line no-unused-vars
 }
 
+interface ColorProperties {
+  background: string;
+  color: string;
+}
+
 /**
  * Overlay that adds a sliding badge to status badges allowing
  * users to quickly create an AI chat message about the badged state.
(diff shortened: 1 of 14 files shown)


pkg/rancher-ai-ui/handlers/hooks/overlay/badge-sliding.ts
```diff
@@ -25,48 +31,111 @@ class BadgeSlidingOverlay extends HooksOverlay {
...
+      g: 255,
+      b: 255
+    };
+
+    const blendedR = Math.round(r * a + blendBg.r * (1 - a));
+    const blendedG = Math.round(g * a + blendBg.g * (1 - a));
+    const blendedB = Math.round(b * a + blendBg.b * (1 - a));
+
+    return `rgb(${ blendedR }, ${ blendedG }, ${ blendedB })`;
+  }
```
marcelofukumoto: **Suggestion**
Cant we programatically "try" to get the correct values? Feels like hardcoding like this is not optimal.

marcelofukumoto (commented):
**Nit**
Just something with the colors, would be better if not hardcoded and retrieved from somewhere else

pkg/rancher-ai-ui/handlers/hooks/overlay/badge-sliding.ts
torchiaf: Right - I pushed a new commit to get the correct background color from the CSS vars defined in Rancher.
For dark theme it's `{ r: 37, g: 40, b: 47 }` != previous hardcoded value `{ r: 40, g: 40, b: 40 }`, so I adapted the e2e as well to reflect the new colors

marcelofukumoto: Great! Only one detail, the colors utils only accetp 6 digits. I guess that is ok, just sharing one Nit that I found.

marcelofukumoto (approved):
LGTM

PR #274, opened 2026-07-29
Explore provisioning clusters

torchiaf:
When the action resource is a CLuster, we can directly redirect to the Cluster Management product, which is built-in and the route is known.

### DIFF
--- a/pkg/rancher-ai-ui/components/message/resource-buttons/resource-context.ts
+++ b/pkg/rancher-ai-ui/components/message/resource-buttons/resource-context.ts
@@ -283,9 +283,9 @@ export function getKnownRoute(actionResource: ActionResource, rancherResource: R
         }
       };
     }
+  default:
+    return null;
   }
-
-  return null;
 }
 
 /**
--- a/pkg/rancher-ai-ui/components/message/resource-buttons/resource-context.ts
+++ b/pkg/rancher-ai-ui/components/message/resource-buttons/resource-context.ts
@@ -252,12 +252,10 @@ export function getDetailLocation(
   actionResource: ActionResource = {},
   rancherResource: RancherResource = {},
 ) {
-  if (rancherResource) {
-    const route = getKnownRoute(actionResource, rancherResource);
+  const route = getKnownRoute(actionResource, rancherResource);
 
-    if (route) {
-      return route;
-    }
+  if (route) {
+    return route;
   }
 
   return buildDetailLocation(store, productName, schema, inStore, actionResource);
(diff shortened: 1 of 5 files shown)


pkg/rancher-ai-ui/components/message/resource-buttons/resource-context.ts
```diff
@@ -219,6 +233,62 @@ export function getProduct(store: Store, schema: Schema | null, resourceType = '
...
+ */
+export function getDetailLocation(
+  store: Store,
+  productName: string,
+  schema: Schema | null,
+  inStore: string,
+  actionResource: ActionResource = {},
+  rancherResource: RancherResource = {},
+) {
+  if (rancherResource) {
```
marcelofukumoto: **Nit**
rancherResource is default to an empty object. So that means here will always be true.

pkg/rancher-ai-ui/components/message/resource-buttons/resource-context.ts
```diff
@@ -219,6 +233,62 @@ export function getProduct(store: Store, schema: Schema | null, resourceType = '
...
+}
+
+/**
+ * Get the route for built-in products, for a given action resource and rancher resource.
+ * @param actionResource the resource returned from the message action
+ * @param rancherResource the resource returned from the rancher API
+ * @returns an object containing the route name and parameters for the resource
+ */
+export function getKnownRoute(actionResource: ActionResource, rancherResource: RancherResource) {
+  switch (rancherResource.type) {
```
marcelofukumoto, line 271: **Question**
Why exactly we are using a switch here? 
There is only one case, right?
I believe we could just use an IF.

pkg/rancher-ai-ui/components/message/resource-buttons/resource-context.ts
torchiaf: Fixed

pkg/rancher-ai-ui/components/message/resource-buttons/resource-context.ts
torchiaf: In the future we may want to add more types here having known routes - adding a comment.

marcelofukumoto: If that is the case, as a **Nit**
Make the return null a default case. I think it is better to always have a default case for a switch.

marcelofukumoto (approved):
LGTM

PR #275, opened 2026-07-30
Add Privacy Notice banner to Local provider

### LINKED ISSUE #194: add disclaimer
We should add a disclaimer for Local LLM 

```
Please be aware that SUSE does not own or control this LLM and is not responsible for its output. 

Your prompts and data will be sent to the provider’s servers. These providers may use your data according to their own privacy policies. 
```

torchiaf:
Fixes: https://github.com/rancher/rancher-ai-ui/issues/194

We are adding the privacy notice to the `Local` provider and fixing the text of `third-party`'s for consistency (the first line is now the same for all providers)

### Screenshots

Local:

image

Others:

image

### DIFF
--- a/pkg/rancher-ai-ui/l10n/en-us.yaml
+++ b/pkg/rancher-ai-ui/l10n/en-us.yaml
@@ -358,8 +358,10 @@ aiConfig:
         banner:
           privacy:
             header: "Privacy Notice:"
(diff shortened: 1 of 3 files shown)


marcelofukumoto (approved):
LGTM

PR #276, opened 2026-07-30
Fix e2e test planning

torchiaf:


marcelofukumoto:
Interesting that this one has a conflict

PR #277, opened 2026-08-03
Helm Chart and k3s version dynamic fetch

torchiaf:
We are now fetching the Rancher Helm chart URL and k3s from `branches-metadata.json` flle from `rancher/dashboard` repo.

The `load-rancher-branches-metadata` fails and blocks the e2e tests when:

- the metadata file name changes
- the values are not found where they are supposed to be in the file structure
- the values are empty or null


marcelofukumoto (approved):
LGTM

Issue #110, opened 2025-12-19
Test extension's releases


torchiaf:
We are now testing the extension running in dev mode.

We want to implement a Jenkins pipeline to install the latest extension version in Rancher in production mode.
This can be done on a daily basis.

PR #280, opened 2026-08-05
Compatibility tests

torchiaf:
Contributes to https://github.com/rancher/rancher-ai-ui/issues/110

Implement a pipeline that runs daily, to test Liz against supported Rancher versions (master, release-2.15, ...)


.github/workflows/rancher-compatibility-tests.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  compatibility:
+    name: "Rancher: branch ${{ matrix.rancher_config.branch }}, tag ${{ matrix.rancher_config.tag }}"
+    strategy:
+      matrix:
+        rancher_config:
+          # Change this list to add/remove Rancher supported versions
+          - { branch: 'master', tag: 'head' }
+          - { branch: 'release-2.15', tag: 'v2.15-head' }
```
marcelofukumoto, line 26: **Suggestion**
For the strategy you want to add a key called:
`fail-fast: false`
If you dont have that, the moment that one fails it will not consider the other one and close as failed.

Maybe you want to add as well the max-parallel. Not sure what is the default for this one but having it explicit can be easier to control

.github/workflows/rancher-compatibility-tests.yaml
```diff
@@ -0,0 +1,33 @@
...
+      run_all_versions:
+        description: 'Run tests for all Rancher versions'
+        required: false
+        type: boolean
+        default: true
+
+jobs:
+  compatibility:
+    name: "Rancher: branch ${{ matrix.rancher_config.branch }}, tag ${{ matrix.rancher_config.tag }}"
+    strategy:
```
marcelofukumoto, line 19: **Nit**
On the rancher/dashboard we normally add this:
    if: github.repository_owner == 'rancher' || vars.ENABLE_AGENTIC_WORKFLOWS == 'true'

One to disable/enable the workflow and to not run on forks. Since it is a cron, would be better to have more control.

torchiaf:
Thanks @marcelofukumoto - I added the suggested changes .
Here is an example of the compatibility jobs run as well: https://github.com/rancher/rancher-ai-ui/actions/runs/31110045714

marcelofukumoto (approved):
LGTM

